PR #70, opened 2015-01-12
mention need to escape % when configuring log file format via configuration file

boegel:
@fgeorgatos, @wpoely86: please review?

PR #71, opened 2015-01-12
mention need to escape % when configuring log file format via configuration file

boegel:
replaces #70 and targets `develop`


boegel:
reviewed in #70

PR #73, opened 2015-01-26
add clear requirements section to installation page, fix pip install prefix

boegel:
@wpoely86, @fgeorgatos: please review

rendered version @ boegel-eb.readthedocs.org/en/develop/Installation.html and http://boegel-eb.readthedocs.org/en/develop/Installation_Alternative.html#alt-inst-methods

### DIFF
--- a/docs/Installation_Alternative.rst
+++ b/docs/Installation_Alternative.rst
@@ -36,13 +36,13 @@ you can use one of the following simple commands:
 
 -  using ``easy_install`` (old tool, but still works)::
 
-       easy_install --prefix /tmp/$USER easybuild
+       easy_install --prefix $HOME/EasyBuild easybuild
 
 -  using ``pip`` (more recent and better installation tool for Python software)::
 
-       pip install --install-option "--prefix=/tmp/$USER" easybuild
+       pip install --install-option "--prefix=$HOME/EasyBuild" easybuild
 
-The ``--prefix /tmp/$USER`` part in these commands allows you to install EasyBuild without admin rights into ``/tmp/$USER``.
+The ``--prefix $HOME/EasyBuild`` part in these commands allows you to install EasyBuild without admin rights into ``$HOME/EasyBuild``.
 
 
 Adjusting ``$PATH`` and ``$PYTHONPATH`` environment variables
@@ -55,8 +55,8 @@ On (most) Linux distributions, the command for doing this is:
 
 .. code:: bash
 
-    export PATH=/tmp/$USER/bin:$PATH
-    export PYTHONPATH=/tmp/$USER/lib/python2.7/site-packages:$PYTHONPATH
+    export PATH=$HOME/EasyBuild/bin:$PATH
+    export PYTHONPATH=$HOME/EasyBuild/lib/python2.7/site-packages:$PYTHONPATH
 
 .. tip::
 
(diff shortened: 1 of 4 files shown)


docs/Installation_Alternative.rst
```diff
@@ -36,13 +36,13 @@ you can use one of the following simple commands:
 
 -  using ``easy_install`` (old tool, but still works)::
 
-       easy_install --prefix /tmp easybuild
+       easy_install --prefix /tmp/$USER easybuild
```
wpoely86: I suggest a more sensible default. Why not `$HOME/opt` ?

docs/Installation_Alternative.rst
```diff
@@ -64,14 +64,14 @@ On (most) Linux distributions, the command for doing this is:
   environment variable for a given installation prefix, you can use the
   following command::
 
-    python -c "import distutils.sysconfig; print distutils.sysconfig.get_python_lib(prefix='/tmp');"
+    python -c "import distutils.sysconfig; print distutils.sysconfig.get_python_lib(prefix='/tmp$USER/');"
```
wpoely86: slash missing between `/tmp` and `$USER`

docs/Installation_Alternative.rst
```diff
@@ -64,14 +64,14 @@ On (most) Linux distributions, the command for doing this is:
...
 
 
 Install with admin rights
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If you do have admin rights on the system where you want to install
-EasyBuild, you can simply omit the ``--prefix /tmp``
+EasyBuild, you can simply omit the ``--prefix /tmp$USER/``
```
wpoely86: same here

docs/Installation_Alternative.rst
```diff
@@ -133,7 +133,7 @@ the easyconfigs package depends on both the framework and easyblocks packages.
 If you do not have ``pip`` or ``easy_install`` available, you can also
 fall back to using the ``setup.py`` script directly::
 
-    python setup.py --prefix /tmp install
+    python setup.py --prefix /tmp$USER install
```
wpoely86: slash

docs/Installation_Alternative.rst
boegel: nice catch!

docs/Installation_Alternative.rst
boegel: I don't think there is a sensible default... I'd like to stay away from suggesting `$HOME`, that's too much of a can of worms with Python (think `--user` which installs in the magic path `$HOME/.local`).

`/tmp/$USER` already beats `/tmp` out of the park, imho

wpoely86: I really don't like `/tmp` for such things. That's only for temporarily files that should not survive reboot. `/var/tmp` might be a slightly better location.
But I would really for go something under `$HOME` ? I've been using `$HOME/opt` since forever without issue? If we use `--prefix`, I don't really see a lot of problems?

boegel: If you feel strongly about it, I'd go with something like `$HOME/eb` or `$HOME/EasyBuild` or somesuch. Anything other than `$HOME/.local` should be fine in terms of Python magic.

wpoely86: That's also fine by me.

I think that if somybody just copy&pastes these commands, they should end up with something that will keep on working.

boegel:
@wpoely86: changed the `/tmp/$USER` to `$HOME/EasyBuild`, please recheck?

docs/Installation_Alternative.rst
```diff
@@ -55,8 +55,8 @@ On (most) Linux distributions, the command for doing this is:
 
 .. code:: bash
 
-    export PATH=/tmp/bin:$PATH
-    export PYTHONPATH=/tmp/lib/python2.7/site-packages:$PYTHONPATH
+    export PATH=$HOME/EasyBuild/bin:$PATH
+    export PYTHONPATH=$HOME/EasyBuild/lib/python2.7/site-packages:$PYTHONPATH
```
wpoely86: what if we have python 2.6...

boegel: see the `tip` below, it mentions how to figure out the path to use

boegel: this is one of the reasons why people should be using the bootstrap ;)

boegel:
Going in, thanks for the thorough review @wpoely86!

PR #91, opened 2015-03-05
add page on no longer supported functionality, document fix_broken_easyconfigs.py script

boegel:
rendered versions available at:
* http://boegel-eb.readthedocs.org/en/develop/No-longer-supported-functionality.html
* http://boegel-eb.readthedocs.org/en/develop/Deprecated-functionality.html
* http://boegel-eb.readthedocs.org/en/develop/Useful-scripts.html

@fgeorgatos, @wpoely86, @rjeschmi: please review?

### DIFF
--- a/docs/Deprecated-functionality.rst
+++ b/docs/Deprecated-functionality.rst
@@ -5,9 +5,37 @@ Deprecated functionality
 
 Some of the functionality that was available in previous EasyBuild versions is *deprecated*, and should no longer be used.
 
-This page describes the :ref:`deprecation policy <deprecation_policy>` in EasyBuild, explains how to easily :ref:`check
-whether you are still relying on deprecated functionality <how_to_check_use_of_deprecated_functionality>`, and provides
-an :ref:`overview of currently deprecated functionality <overview_deprecated>`, together with available alternatives.
+Support for using this functionality will be dropped in a future version of EasyBuild (see :ref:`no_longer_supported`
+for more information about functionality for which support has already been dropped).
+
+This page provides an :ref:`overview of currently deprecated functionality <overview_deprecated>` together with
+available alternatives, describes the :ref:`EasyBuild deprecation policy <deprecation_policy>`, and explains how
+to easily :ref:`check whether you are still relying on deprecated functionality
+<how_to_check_use_of_deprecated_functionality>`.
+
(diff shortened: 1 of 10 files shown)


wpoely86:
looks fine

rjeschmi:
For future edits: It feels like "no longer supported" could be replaced with "removed" just to make it flow better (deprecated => removed).

docs/Deprecated-functionality.rst
```diff
@@ -5,9 +5,39 @@ Deprecated functionality
 
 Some of the functionality that was available in previous EasyBuild versions is *deprecated*, and should no longer be used.
 
-This page describes the :ref:`deprecation policy <deprecation_policy>` in EasyBuild, explains how to easily :ref:`check
-whether you are still relying on deprecated functionality <how_to_check_use_of_deprecated_functionality>`, and provides
-an :ref:`overview of currently deprecated functionality <overview_deprecated>`, together with available alternatives.
+Support for using this functionality will be dropped in a future version of EasyBuild (see :ref:`no_longer_supported`
```
rjeschmi: remove: "support for using"

consider using "removed" instead of "dropped"

leaving: "This functionality will be dropped in a future version of EasyBuild"

docs/Deprecated-functionality.rst
```diff
@@ -5,9 +5,39 @@ Deprecated functionality
...
+for more information about functionality for which support has already been dropped).
+
+This page:
+* provides an :ref:`overview of currently deprecated functionality <overview_deprecated>` together with
```
rjeschmi: IMPORTANT: Not rendering as a list on the rendered version

pforai:
I wasn't happy at first with the new structure and page titles because I was thinking that they are too similar and because of that a bit misleading, but after thinking about how to do it 'better' I couldn't come up with anything that makes more sense. So good to go as it is now, I'd say.

rjeschmi:
@pforai I find that some of the titles are a bit long in any case.

Nothing else that blocks this PR, but we might use the hackathons as a bit of a focus group, especially with newer users...

boegel:
@rjeschmi: excellent suggestion w.r.t. using 'removed functionality', feels better indeed

boegel:
going in, thanks for the feedback everyone!

PR #128, opened 2015-07-03
fix documentation w.r.t. easyblock parameter, no more fallback to ConfigureMake

boegel:
This should've been changed ages ago, since the fallback has been disabled since EB v2.0...

@wpoely86: please review?

### DIFF
--- a/docs/Changelog.rst
+++ b/docs/Changelog.rst
@@ -3,6 +3,9 @@
 Changelog for EasyBuild documentation
 -------------------------------------
 
+* **release 20150703.01** (`July 3rd 2015`):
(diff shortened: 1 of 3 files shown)


rjeschmi:
It looks generally fine, but I should check the context. It is somewhat verbose and I wonder if it couldn't be simplified...

boegel:
@rjeschmi: rendered version at http://boegel-eb.readthedocs.org/en/easyblock/Writing_easyconfig_files.html#easyblock-specification doesn't look too verbose to me?

rjeschmi:
It looks good (and better)

boegel:
Thanks for the review @rjeschmi!

PR #138, opened 2015-08-06
Script for generating API documentation

Caylo:
Also includes the generated docs

For prettifying, some things in the docstrings will have to be changed, e.g @param and @author tags

For author tags, use field lists (`:author: Name`, see http://sphinx-doc.org/markup/misc.html#file-wide-metadata)

See also: http://sphinx-doc.org/domains.html#the-python-domain (especially the "into fields list" section)


boegel:
replaced by #139

PR #139, opened 2015-08-06
Script to generate API documentation

Caylo:
Also includes the generated docs. The API docs use `automodule` which means they will always be built from the last released version.

For prettifying, some things in the docstrings will have to be changed, e.g @param and @author tags

For author tags, use field lists (`:author: Name`, see http://sphinx-doc.org/markup/misc.html#file-wide-metadata)

See also: http://sphinx-doc.org/domains.html#the-python-domain (especially the "into fields list" section)


boegel:
good to go, thanks @Caylo!

PR #146, opened 2015-08-13
add documentation on options/parameters for controlling permissions (WIP)

boegel:
fueled by https://github.com/hpcugent/easybuild-framework/issues/1351

@ananke, @ocaisa, @gppezzi: I'd love to get some help with this from you, since you have experience and probably also internal documentation on this already?

### DIFF
--- a/docs/Controlling_permissions_of_installations.rst
+++ b/docs/Controlling_permissions_of_installations.rst
@@ -1,0 +1,60 @@
+.. _controlling_permissions:
+
+Controlling permissions of installations
+========================================
+
+Various configuration options and easyconfig parameters are available that control (group) ownership and permissions of
+software installations performed with EasyBuild.
+
+These can be used to protect installations of licensed software, or to make sure that installations can be
+modified/removed by other users which can be interesting in case multiple users are performing installations using
+EasyBuild to a shared filesystem.
+
+.. note::
+
+  These options and parameters only affect the permissions and (group) ownership of the actual software installations
+  themselves, not of the module files that are generated for them or other files being generated/copied (e.g. the
+  easyconfig file that is copied to the easyconfigs repository).
+
+Configuration options
+---------------------
+
+These configuration options are *global*, in the sense that they apply to all software installations being performed
+when they are set.
+
+We'll refer to them via their respective command line option, but just like all other configuration option these can
+also be set via an EasyBuild configuration file or via an ``$EASYBUILD*`` environment variable
+(see :ref:`configuration_consistency`).
+
+Group ownership (``--group``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Force group-writeable installations (``--group-writable-installdir``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Forced read-only installations (``--read-only-installdir``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+GID bit (``--set-gid-bit``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sticky bit (``--sticky-bit``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+File creation mode mask (``--umask``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+Easyconfig parameters
+---------------------
+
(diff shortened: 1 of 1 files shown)


docs/Controlling_permissions_of_installations.rst
```diff
@@ -0,0 +1,60 @@
...
+
+These configuration options are *global*, in the sense that they apply to all software installations being performed
+when they are set.
+
+We'll refer to them via their respective command line option, but just like all other configuration option these can
+also be set via an EasyBuild configuration file or via an ``$EASYBUILD*`` environment variable
+(see :ref:`configuration_consistency`).
+
+Group ownership (``--group``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ocaisa: This is probably the one that needs a little explaining since it would be unusual to use this as a configure option, something like ```Name of the user group for which the software should be available, typical usage is to restrict access to a particular unix group due to the use of licensed software```

docs/Controlling_permissions_of_installations.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Force group-writeable installations (``--group-writable-installdir``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Forced read-only installations (``--read-only-installdir``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+GID bit (``--set-gid-bit``)
```
ocaisa: When using this as a site-wide configuration option, this will force the entire installation tree to belong to the same group as the original root directory. This is necessary when using when multiple users are responsible for installations and the unix group that is common to them is not their primary group.

docs/Controlling_permissions_of_installations.rst
```diff
@@ -0,0 +1,60 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sticky bit (``--sticky-bit``)
```
ocaisa: The sticky bit ensures that (unless overridden with the ```--group-writable-installdir``` option) only the installer of a piece of software can change it's name (so someone cannot force a reinstall by simply renaming the installation directory).

docs/Controlling_permissions_of_installations.rst
```diff
@@ -0,0 +1,60 @@
...
+We'll refer to them via their respective command line option, but just like all other configuration option these can
+also be set via an EasyBuild configuration file or via an ``$EASYBUILD*`` environment variable
+(see :ref:`configuration_consistency`).
+
+Group ownership (``--group``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Force group-writeable installations (``--group-writable-installdir``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
ocaisa: Regardless of the value of the umask configuration option, the default is to create only user-writable installations (only the original installer may re-install a package). This setting will override this default and create group-writable installations.

docs/Controlling_permissions_of_installations.rst
```diff
@@ -0,0 +1,60 @@
...
+GID bit (``--set-gid-bit``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sticky bit (``--sticky-bit``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+File creation mode mask (``--umask``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
ocaisa: Override the default umask settings. For example, when multiple users are responsible for installations (and these users belong to a common group) then you would set the umask to 002

docs/Controlling_permissions_of_installations.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Easyconfig parameters
+---------------------
+
+These easyconfig parameters can be defined in easyconfig files to control ownership/permission aspects for the
+particular installation that results from using that easyconfig file.
+
+Group ownership (``group``)
```
ocaisa: This option restricts read access to a particular unix group. Typical usage is for the case of licensed software. The installer must also belong to this group and, if necessary, have executed ```newgrp [group]``` prior to attempting the installation.

docs/Controlling_permissions_of_installations.rst
Flamefire: This needs a bit more information. E.g. mention that this **removes** some permissions if they are set. Like "002" removes the read permissions for "others" (users except the owner and the group)

PR #155, opened 2015-09-23
Update Using_the_EasyBuild_command_line.rst

cfenoy:
Small clarification about search path for dependencies


cfenoy:
target wrong branh, sorry

PR #161, opened 2015-10-07
documentation for .yeb format

Caylo:
Compiled version: http://caylo-easybuild.readthedocs.org/en/yeb-docs/Writing_yeb_easyconfig_files.html


rjeschmi:
Did we get the docs for the yeb options anywhere yet?

boegel:
What do you mean with yeb options?

rjeschmi:
Maybe what I'm looking for is pre-release docs. We could put them somewhere else (ie, checkout feature branch, use --experimental etc).

Also, the example in the docs seems to be slightly wrong (easyblock shouldn't be specified, or should be bzip I think, since there is no configure script used for the sources)

rjeschmi:
Also the escaping in the docs for *SOURCE_TAR_GZ (shows up as \*SOURCE_TAR_GZ).

Is * a preference or something that makes sense to YAML? It is going to be a bit of a pain long term I imagine (for the same reason the docs are hard to escape).

boegel:
We currently don't have documentation for pre-release stuff, but we can add instructions here on how to start playing around with this.
We can even provide an easyconfig for to install EasyBuild 2.x.0dev with .yeb support included.

rjeschmi:
that might be nice. This is a really cool feature. Definitely looking forward to this moving forward.

boegel:
And you're right about the `easyblock` thing, we took a shortcut there to use a `bzip2` as a test case, to avoid requiring the `EB_bzip2` easyblock. I guess we should be using another one.

rjeschmi:
Maybe we should create a feature branch of easyconfigs with some interesting examples merged?

boegel:
The `*SOURCE_TAR_GZ` is standard YAML, see http://www.yaml.org/spec/1.2/spec.html#alias//
The definition of the alias with `&SOURCE_TAR_GZ` is injected by `eb` right before the `.yeb` file is parsed.

It's the best we found so far to as equivalent to the `SOURCE_TAR_GZ` constants we now have, since there's no discrimination between strings and variables in YAML otherwise.

boegel:
A feature branch with easyconfigs makes sense to me. Question is how we should organise things there. Do we simply mix `.eb` and `.yeb` files in the same `<software>` directory?

rjeschmi:
Yes, I think so about mixing. Though I think we expect to have only one .yeb per software (though we may allow inheritance of some sort)

docs/Writing_yeb_easyconfig_files.rst
```diff
@@ -0,0 +1,206 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code::
+
+    %YAML 1.2
+    ---
+    easyblock: ConfigureMake
+
+    name: gzip
+    version: '1.6'  # FIXME bug: quotes are required here to make sure this is parsed a a string, not a floating point value
```
boegel: @Caylo: please look into this, make very sure somehow that `version` is string after parsing...

It's probably time we define the expected type for each of the easyconfig parameters...

Also, please include this `gzip.yeb` in the test easyconfigs, and include it in the tests for `.yeb`.

docs/Writing_yeb_easyconfig_files.rst
```diff
@@ -0,0 +1,206 @@
...
+    sources: ['%(name)s-%(version)s.tar.gz']  # equivalent with [*SOURCE_TAR_GZ]
+
+See also :ref:`easyconfig_param_templates`.
+
+.. _easyconfig_yeb_format_syntax_dependencies:
+
+Dependencies
+~~~~~~~~~~~~
+
+*(WORK IN PROGRESS)*
```
boegel: @Caylo: this is all yours to figure out ;-)

docs/Writing_yeb_easyconfig_files.rst
```diff
@@ -0,0 +1,245 @@
...
+
+As string concatenation is not supported in YAML, we defined a !join operator. It takes a list of values to concatenate
+as argument. This list can contain both hard strings and variables.
+Example::
+
+    !join [foo, bar]  # returns 'foobar'
+    vars:
+        - &f foo
+        - &b bar
+    !join [f, bar]  # returns 'foobar'
```
boegel: this works with `f`, no `*f` required?

docs/Writing_yeb_easyconfig_files.rst
```diff
@@ -0,0 +1,245 @@
...
+
+Dependencies
+~~~~~~~~~~~~
+
+*(WORK IN PROGRESS)*
+Variables can be defined using standard YAML anchors (using a '&'). These are later referenced using an asterisk (*).
+Example::
+    vars:
+        - &compver 2015.3.187
+        - &gccsuff -GNU-4.9.3-2.25
```
boegel: I'll rework this a bit, this section should be on dependencies specifically, using internal variables is more generic, and should be described in a separate section.

rjeschmi:
Add this feature to the list of experimental features in:

:ref:`experimental_features

docs/Writing_yeb_easyconfig_files.rst
```diff
@@ -0,0 +1,359 @@
...
+    using them currently requires enabling the use of experimental features (``--experimental``),
+    see also :ref:`experimental_features` .
+
+    An up-to-date overview of current progress on support for ``.yeb`` easyconfigs is available at
+    https://github.com/hpcugent/easybuild-framework/issues/1407.
+
+Useful links:
+
+* YAML syntax specification: http://www.yaml.org/spec/1.2/spec.html
+
```
rjeschmi: Is PyYAML a 1.2 parser? They advertise as 1.1 (though I imagine there are some extensions)

boegel: Hmm, good point.

Although, PyYAML is not complaining when we're using `%YAML 1.2` as a header.

rjeschmi:
Merge it as is, things look fine and generally good. We'll work with it and update as we learn more.

boegel:
@rjeschmi: makes sense, let's do that indeed, thanks for the review!

PR #171, opened 2015-11-27
Add custom source of distribute 0.6.49 for downloading

wpoely86:
The PR for which this is need can be found at
https://github.com/hpcugent/easybuild-framework/pull/1490

This is the source of distribute 0.6.49 with the file
`_markerlib/markers.py` replaced by the 0.6 version of
markerlib which can be found at
https://pypi.python.org/pypi/markerlib/0.6.0


wpoely86:
@boegel please merge so I can retest the PR

boegel:
@wpoely86: hmm, can we rename the tarball to indicate it's patched version, while still allowing it to be picked up?

maybe something like `files/distribute-0.6.49-1.tar.gz` (or, if possible, something more descriptive than the extra `-1`)?

wpoely86:
@boegel that's more patching in the bootstrap scripts of distribute...

boegel:
@wpoely86: the tarball name is hardcoded? we're already patching it, so shouldn't be that hard?

wpoely86:
It's possible
but there might be unexpected fallout...

wpoely86:
@boegel done, please merge

Issue #173, opened 2015-12-09
readthedocs - default easyblock (deprecated)


JackPerdue:
At: https://easybuild.readthedocs.org/en/latest/Concepts_and_Terminology.html#easyblocks

" In case an easyblock specification is not provided and no (software-specific) easyblock matching the software name could be found, a fallback mechanism will resort to using the generic ConfigureMake easyblock, which implements the common configure / make / make install procedure."

That behaviour was deprecated in 2.x (2.0?).

boegel:
Nice catch @JackPerdue, I'll get that sorted.

boegel:
see #174

PR #174, opened 2015-12-09
fix outdated fallback to ConfigureMake being mentioned in concepts page

boegel:
fix for #173


boegel:
Thanks for reporting this @JackPerdue!

PR #197, opened 2016-02-02
fix typo in docs

ocaisa:


docs/Extended_dry_run.rst
```diff
@@ -26,7 +26,7 @@ produces.
 
 .. _extended_dry_run_notes_differences:
 
-Build/install procedure reported by dryn run may be (slightly) different
+Build/install procedure reported by dry run may be (slightly) different
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
boegel: @ocaisa: I think you better match the length here with the above, otherwise the formatting may go all weird

PR #199, opened 2016-02-13
update docs on GitHub integration, mention --new-pr/--update-pr

boegel:
preview at http://boegel-eb.readthedocs.org/en/new_update_pr/Integration_with_GitHub.html


boegel:
@wpoely86, @pescobar, @fgeorgatos, @rjeschmi I finally got around to completing this... Please review?

pescobar:
why not enable `--test-report-env-filter` by default and enable it only in case it's needed and you know what you are doing?

maybe it's worth mentioning that a recent python-keyring version is needed? I remember I had issues with this but I don't rember the specific version I had to use...

otherwise, lgtm

boegel:
@pescobar well, I guess we could change the default to the filter being mentioned in the documentation now, sure... @geimer: thoughts on this, since you cooked this up?

besserox:
@boegel  Quick feedback about the `--git-working-dirs-path` option:

It says *location of your Git working directories*, which one? easyconfig, easyblock, ... ? only one? multiples? all of them? 

I think this could be clarified.

docs/Integration_with_GitHub.rst
```diff
@@ -121,3 +276,361 @@ in that same pull request for netCDF, WRF, ...::
...
+To create a new pull request, the ``--new-pr`` command line option can be used, provided that the necessary
+requirements are fullfilled (see :ref:`github_requirements`).
+
+In its simplest form, you just provide the location of the file(s) that you want to include in the pull request::
+
+    $ eb --new-pr test.eb --experimental
+
+This takes care of all the steps required to make a contribution, i.e.:
+
+* set up a working copy of the relevant EasyBuild repository (e.g., ``easybuild-easyconfigs``)
```
geimer: This is a bit unclear to me.  From the example, it seems that only easyconfigs can be handles by this option.  But with the information about `--pr-target-repo` below, I get the impression that also framework/easyblock PRs can be handled.  However, in this case, the question is how the automatic renaming/moving works.

geimer:
@boegel: I don't think I was involved in this stuff... But anyway, I did a quick review and in general it looks reasonable to me.

docs/Integration_with_GitHub.rst
boegel: @geimer For now, in the experimental stage, `--new-pr` and `--update-pr` only support easyconfigs, but the intention is to extend this to easyblocks and framework too.

For easyblocks, we can derive the correct name & location as well. For framework, we will only be able to support patching of existing modules, but that should cover 95% of all uses.

The documentation is written with that in mind, and so is the `--pr-target-repo` configure option.

boegel:
@geimer I was pinging you w.r.t. the value for `--test-report-env-filter` in particular, I think you came up with the filter that is mentioned in the documentation?

@pescobar's suggestion was to make that the default...

geimer:
@boegel: I don't think so.  A `--test-report-env-filter` example was already in the wiki when I started using EasyBuild...

boegel:
@geimer hmm, ok, sorry for the noise then, I must be misremembering...

boegel:
going in, thanks for the feedback everyone!

PR #200, opened 2016-02-14
add new docs section on --show-config

boegel:
preview at http://boegel-eb.readthedocs.org/en/show_config/Configuration.html

### DIFF
--- a/docs/Configuration.rst
+++ b/docs/Configuration.rst
@@ -1,7 +1,3 @@
-
-.. toctree::
-     :maxdepth: 1
-
 .. _configuring_easybuild:
 
 Configuring EasyBuild
@@ -9,6 +5,10 @@ Configuring EasyBuild
 
 This page discusses the recommended style of configuring
 EasyBuild, which is supported since EasyBuild v1.3.0.
+
+.. contents::
+    :depth: 2
+    :backlinks: none
 
 .. _configuration_types:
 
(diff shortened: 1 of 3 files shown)


boegel:
@wpoely86: please review?

docs/Configuration.rst
```diff
@@ -287,6 +287,57 @@ Examples (more below):
...
+    # (C: command line argument, D: default value, E: environment variable, F: configuration file)
+    #
+    buildpath      (F) = /tmp/eb-build
+    installpath    (C) = /Users/example/apps
+    job-cores      (C) = 4
+    modules-tool   (E) = Lmod
+    optarch        (E) = ''
+    repositorypath (D) = /Users/example/.local/easybuild/ebfiles_repo
+    robot-paths    (D) = /Users/example/easybuild-easyconfigs/easybuild/easyconfigs
+    sourcepath     (D) = /Users/example/.local/easybuild/sources
```
fgeorgatos: all good, useful feature.

an idea that comes to mind: it would be nice, if the output corresponded directly with shell variables, fi.:
`EASYBUILD_MODULES_TOOL=Lmod` etc; OTOH, this just one of the many ways to tune things...

boegel: I think the output should be uniform in any case, and clearly readable.

The way it is printed now is very readable imho, a one-to-one mapping with entries in configuration *and* very close to the corresponding command line  arguments (just prefix with `--`).

I feel the mapping to the corresponding environment variable is trivial enough (upper case, prefix with `EASYBUILD_` and translate `-` to `_`) not to clutter the output with `EASYBUILD_` everywhere...

wpoely86:
lgtm

boegel:
https://github.com/hpcugent/easybuild-framework/pull/1611 is merged, so good to go, thanks for the feedback @fgeorgatos and @wpoely86!

PR #212, opened 2016-04-19
DOC: instructions to update an existing install

andreas-h:
following an [e-mail](https://lists.ugent.be/wws/arc/easybuild/2016-04/msg00023.html) from @boegel

### DIFF
--- a/docs/Installation.rst
+++ b/docs/Installation.rst
@@ -318,6 +318,18 @@ Example usage, with the relevant output at the start of stage 1 of the bootstrap
           Source tarballs for all three EasyBuild components *must* be provided when ``$EASYBUILD_BOOTSTRAP_SOURCEPATH``
           is defined, however.
 
+.. _updating:
+
+Updating an existing EasyBuild installation
+-------------------------------------------
+
+To upgrade to a newer EasyBuild version (say, *x.y.z*) than the one currently installed (say, *a.b.c*), there are several options:
+
+     (i) (re)bootstrap EasyBuild to obtain an EasyBuild/x.y.z module to load, using the instructions above, see :ref:`bootstrapping`.
+     (ii) install EasyBuild x.y.z with a previous version of EasyBuild, using the easyconfig file available `on the develop branch at Github <https://github.com/hpcugent/easybuild-easyconfigs/tree/develop/easybuild/easyconfigs/e/EasyBuild>`__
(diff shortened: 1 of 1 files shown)


docs/Installation.rst
```diff
@@ -318,6 +318,18 @@ Example usage, with the relevant output at the start of stage 1 of the bootstrap
...
           is defined, however.
 
+.. _updating:
+
+Updating an existing EasyBuild installation
+-------------------------------------------
+
+To upgrade to a newer EasyBuild version (say, *x.y.z*) than the one currently installed (say, *a.b.c*), there are several options:
+
+     (i) (re)bootstrap EasyBuild to obtain an EasyBuild/x.y.z module to load, using the instructions above, see :ref:`bootstrapping`.
```
boegel: @andreas-h please use backticks for make the module name 'jump out', i.e.:

```
`EasyBuild/x.y.z`
```

docs/Installation.rst
```diff
@@ -318,6 +318,18 @@ Example usage, with the relevant output at the start of stage 1 of the bootstrap
           Source tarballs for all three EasyBuild components *must* be provided when ``$EASYBUILD_BOOTSTRAP_SOURCEPATH``
           is defined, however.
 
+.. _updating:
+
+Updating an existing EasyBuild installation
+-------------------------------------------
+
+To upgrade to a newer EasyBuild version (say, *x.y.z*) than the one currently installed (say, *a.b.c*), there are several options:
```
boegel: rather than using `*x.y.z*`, you can use `|version|`, which will be replaced with the latest EasyBuild version (as defined in `conf.py`)

andreas-h:
Mhh, @boegel, if I use backticks for

    `EasyBuild/x.y.z`

I'd rather not use the `|version|` substitution, because then the user would see the result of `|version|` in some places, and x.y.z in the module name (or is it possible to somehow use a substitution within the backticks?  I couldn't figure out how), which would be confusing.  Do you agree?

boegel:
@andreas-h Yes, you're right, but then I would opt for using *EasyBuild/|version|*, where `|version|` should still get replace properly.

Showing how to upgrade to the latest version makes sense, while using `x.y.z` may make things less clear; people tend to only half read documentation, and look for patterns instead :)

boegel:
merged via #215, thanks @andreas-h!

PR #233, opened 2016-07-05
add toolchain-opts to documentation script

Caylo:


docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,9 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
 echo "*(see also* \`\`eb --list-toolchains\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchains.rst
 echo >> $VERSION_SPECIFIC_DIR/toolchains.rst
 eb --list-toolchains --output-format rst >> $VERSION_SPECIFIC_DIR/toolchains.rst
+
+#  available toolchain options
+echo ".. _avail_toolchain_opts:" > $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo "*(see also* \`\`eb --avail-toolchain-opts <tcname>\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+eb --avail-toolchain-opts intel --output-format rst >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
```
boegel: on 2nd thought, we should probably generate the available toolchain options for *each* of the supported toolchains here, with proper subsections, etc?

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,15 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
...
 eb --list-toolchains --output-format rst >> $VERSION_SPECIFIC_DIR/toolchains.rst
+
+#  available toolchain options
+echo ".. _avail_toolchain_opts:" > $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo "*(see also* \`\`eb --avail-toolchain-opts <tcname>\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+IFS=$'\n'
+for i in `eb --list-toolchains | sed 1d`
+do
+    tc=`echo $i | cut -d ':' -f1 | sed 's/[ \t]//g'`
```
boegel: seems like like this needs `sed 's/[ \t]*//g'` ?

```
ERROR: Failed to parse configuration options: "Couldn't find toolchain: '\tClangGCC'. To see available toolchains, use --list-toolchains"
```

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,15 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
...
+
+#  available toolchain options
+echo ".. _avail_toolchain_opts:" > $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo "*(see also* \`\`eb --avail-toolchain-opts <tcname>\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+IFS=$'\n'
+for i in `eb --list-toolchains | sed 1d`
+do
+    tc=`echo $i | cut -d ':' -f1`
+    eb --avail-toolchain-opts ${tc//[[:blank:]]/} --output-format rst >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
```
boegel: ok, this works now (cfr. http://boegel-eb.readthedocs.io/en/test/version-specific/toolchain_opts.html), but two requests:

* sort toolchains alphabtically
* can we add quicklinks for each of the toolchains on the top of the page? that'll involve inserting labels for each toolchain...

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,27 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
...
+echo ".. _avail_toolchain_opts:" > $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo "*(see also* \`\`eb --avail-toolchain-opts <tcname>\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+IFS=$'\n'
+
+for i in `eb --list-toolchains | sed 1d | sort`
+do
+    tc=`echo $i | cut -d ':' -f1`
+    echo "- ${tc//[[:blank:]]/}_" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
```
boegel: I wouldn't make this a list, since this leads to a very long line on top of the page, cfr. http://boegel-eb.readthedocs.io/en/test/version-specific/toolchain_opts.html

I'd make this a single line instead, like we do on http://boegel-eb.readthedocs.io/en/test/version-specific/generic_easyblocks.html

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,27 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
...
+done
+
+
+for i in `eb --list-toolchains | sed 1d | sort`
```
boegel: try to run this only once (but it's not a big issue)

boegel:
@Caylo: `ccache` only supports C/C++, not Fortran, but there are alternatives available, for example http://people.irisa.fr/Edouard.Canot/f90cache

So, `--use-ccache` should probably be renamed to `--use-compiler-cache`, and for Fortran90 we should use `f90cache` instead.

Do note that it works differently: rather than simply symlinking the original compiler commands, you need to prefix them, so you'll need to create a wrapper script that does this...

boegel:
@Caylo just ran into this while playing around with this WIP branch, please fix? :)

```
ERROR: Traceback (most recent call last):
  File "/data/EasyBuild/easybuild-framework/easybuild/main.py", line 114, in build_and_install_software
    (ec_res['success'], app_log, err) = build_and_install_one(ec, init_env)
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 2363, in build_and_install_one
    result = app.run_all_steps(run_test_cases=run_test_cases)
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 2279, in run_all_steps
    self.run_step(step_name, step_methods)
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 2158, in run_step
    step_method(self)()
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/framework/easyblock.py", line 1554, in prepare_step
    self.toolchain.prepare(self.cfg['onlytcmod'], silent=self.silent)
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/tools/toolchain/toolchain.py", line 647, in prepare
    self.symlink_compilers(ccache)
  File "/home/kehoste/EasyBuild/easybuild-framework/easybuild/tools/toolchain/toolchain.py", line 602, in symlink_compilers
    os.symlink(path, os.path.join(tmpdir, comp))
OSError: [Errno 17] File exists
```

boegel:
Oh, crap, wong PR... >_<

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,36 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
...
+    echo "${tc//[[:blank:]]/}" >> temp_tc
+done
+
+echo -n `cat temp_tc | sed -n 1p`"_" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+
+cat temp_tc | sed 1d | while read line
+do
+    echo -n " - ""$line""_" >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+done
+
```
boegel: @Caylo you need to emit an empty line here with `echo` (without arguments, otherwise you break the first anchor, cfr. http://boegel-eb.readthedocs.io/en/test/version-specific/toolchain_opts.html

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,36 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
...
+
+cat temp_tc | sed 1d | while read line
```
boegel: the `sed 1d` should be dropped here?

docs/scripts/update-all-docs.sh
```diff
@@ -111,3 +111,36 @@ echo ".. _list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
 echo "*(see also* \`\`eb --list-toolchains\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchains.rst
 echo >> $VERSION_SPECIFIC_DIR/toolchains.rst
 eb --list-toolchains --output-format rst >> $VERSION_SPECIFIC_DIR/toolchains.rst
+
+#  available toolchain options
+echo ".. _avail_toolchain_opts:" > $VERSION_SPECIFIC_DIR/toolchain_opts.rst
+echo >> $VERSION_SPECIFIC_DIR/toolchain_opts.rst
```
boegel: please also add a page title above the toc, `Available toolchain options (by toolchain)`

boegel:
Looking great now, thanks @Caylo!

PR #234, opened 2016-07-05
add docs for filtering test suites

Caylo:


docs/Unit-tests.rst
```diff
@@ -182,6 +182,20 @@ To only run the subgroup of tests for ``filetools``::
 
     python -m test.framework.filetools
 
+Subsuites can be filtered on method names. Simply add the string to filter on to the test command. For example, to run
+only the tests containing the word ``load`` in the subgroup ``modules``, run ::
+
+    python -m test.framework.modules load
```
boegel: use `$` to discriminate the command from the output, and include the *full* output (it's short enough)

docs/Unit-tests.rst
```diff
@@ -182,6 +182,20 @@ To only run the subgroup of tests for ``filetools``::
 
     python -m test.framework.filetools
 
+Subsuites can be filtered on method names. Simply add the string to filter on to the test command. For example, to run
```
boegel: `Since EasyBuild v2.8.2, tests can be filtered by name.`

(don't mention subsuites, since the filtering also works for the full suite :))

boegel:
thanks @Caylo!

PR #242, opened 2016-08-02
add API to index page

Caylo:
result http://caylo-easybuild.readthedocs.io/en/api-link/#api


docs/index.rst
```diff
@@ -78,6 +78,16 @@ Other topics
     Deprecated-functionality
     Removed-functionality
 
+API
+===
```
boegel: @Caylo no need for a separate section, just include in section above?

boegel:
Thanks @Caylo!

PR #257, opened 2016-10-10
add page & refs for EasyBuild demos

boegel:
see http://boegel-eb.readthedocs.io/en/asciinema/demos/


fgeorgatos:
lgtm. fun to see the demos, too.

Somewhere on the first demo I noticed: `version it if it` typo likely...

boegel:
@fgeorgatos nice catch!

The script to generate the demo via a patched `asciinema` are now hosted in https://github.com/boegel/eb-asciinema, but I should probably move them here to keep things together...

I'll look into fixing the typo, thanks for the feedback!

boegel:
Thanks for the feedback @fgeorgatos and @rjeschmi!

PR #266, opened 2016-10-24
fix broken ref + clarify compiler flags used for --optarch=GENERIC

boegel:


boegel:
Thanks for pointing this out @geimer!

PR #288, opened 2016-12-12
conda/anaconda docs

jerowe:
Added in a few docs for anaconda and conda.

Thoughts? Opinions?

### DIFF
--- a/docs/Anaconda_easyblocks.rst
+++ b/docs/Anaconda_easyblocks.rst
@@ -1,0 +1,82 @@
+.. _anaconda:
+
+Anaconda
+=========
+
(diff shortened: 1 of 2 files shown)


jerowe:
@boegel , what do you think about these docs?

boegel:
@jerowe the info in `Conda.rst` is basically what I would like to see documented; I'm not sure if we should also add specific documentation on installing Anaconda with EasyBuild, especially not on a separate page, since it's not different from installing anything else with EasyBuild...

jerowe:
oooo I just noticed this was still open. What should I change? Just keep the `Conda.rst` ?

boegel:
@jerowe Yes, I think that's sufficient.

PR #289, opened 2016-12-18
document need to download source tarballs for vsc-install & vsc-base from PyPI

boegel:
preview at http://boegel-eb.readthedocs.io/en/docs_offline_install/Installation.html#offline-bootstrapping-using-supplied-source-tarballs


boegel:
Nice catch, thanks for the suggestion @arif-ali! Done in acdfa72.

PR #294, opened 2017-01-27
Optarch documentation updated

damianam:


docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,36 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+
+The syntax is ``--optarch=Intel:xHost;GCC:-march=native``. As in the simple cases, EasyBuild adds one ``-`` to the flags
+specified, so the flags passed to the compilers in this case are ``-xHost`` and ``--march=native``, respectively.
+
+Additionally, ``GENERIC`` is also supported on a compiler basis, allowing to specify a generic compilation for the desired
+compilers. This is an example of this usage: ``--optarch=Intel:xHost;GCC:GENERIC``. Of course, this is supported just for
+compiler toolchains that recognize ``GENERIC``.
+
+The options for each compiler are set independently. That means that if the GCC toolchain is used, but the only compiler
+specified is Intel (for example with ``--optarch=Intel:xCORE-AVX2``), then the GCC toolchain will ignore that option and
```
boegel: format `Intel` as code?

docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,36 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+behave as ``--optarch`` was never specified.
```
boegel: `s/never/not/`

docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,36 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+
+The compiler name is determined by the ``COMPILER_FAMILY`` constant in the compiler class. At the moment, these are the
+supported names:
+
+* ``GCC``
+* ``Intel``
+* ``PGI``
+* ``Clang``
+* ``IBMXL``
+* ``DUMMY``
```
boegel: I wouldn't mention `DUMMY`, since specifying flags for a `dummy` toolchain doesn't make any sense (they'll never be used)

Also, I think it's actually `ClangGCC` rather than `Clang` (`easybuild/toolchains/compiler/clang.py` doesn't define `COMPILER_FAMILY`).

Which clearly shows we need an option in `eb` to dump the known compiler names, and then mention here that `GCC`, `Intel`, etc. are examples of compiler names and that you can get a full updated list using `eb --avail-compiler-names` (so this documentation doesn't get outdated)

damianam: I partly disagree with the need for a new option. I see your point but finding the compiler family names is trivial (and they are pretty much what you would expect anyway). We can add another option to EasyBuild, but I am afraid that adding trivial options will just add noise to the already extensive list of options that EasyBuild has. `--avail-compiler-names` will be handy for a reduced amount of people, but all of us take a look at the output of `--help` once in a while. Polluting the output that we all read might not be a great idea.

And clang has now a `COMPILER_FAMILY`, called `Clang` :-P

docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,38 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+Setting architecture flags for different compilers via ``--optarch=<compiler:flags>;<compiler:flags>``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Starting with version 3.1.0, EasyBuild supports specifying architecture flags on a per-compiler basis. This enables to
+"set and forget" the ``--optarch`` option for your compilers of interest, as opposed to change it depending on the
+compiler used on the packages to be installed.
+
+The syntax is ``<compiler:flags>;<compiler:flags>``, where ``:`` separates the compiler name from the compiler flags,
+and ``;`` separates different compilers. This is an example for the Intel and GCC compilers:
+``--optarch=Intel:xHost;GCC:'march=x86-64 -mtune=generic'``. As in the simple cases, EasyBuild adds one ``-`` to the
```
boegel: only using quotes for `GCC` screws things up, try `eb --optarch=Intel:xHost;GCC:'march=x86-64 -mtune=generic' --show-config`; you should use this instead:

```
``--optarch='Intel:xHost;GCC:march=x86-64 -mtune=generic'``
```

docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,38 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+flags specified, so the flags passed to the Intel and GCC compilers in this case are ``-xHost`` and
+``-march=x86-64 -mtune=generic``. Please note the quotes to escape the space in the GCC flags.
+
+Additionally, ``GENERIC`` is also supported on a compiler basis, allowing to specify a generic compilation for the
+desired compilers. This is an example of this usage: ``--optarch=Intel:xHost;GCC:GENERIC``. Of course, this is
+supported just for compiler toolchains that recognize ``GENERIC``.
+
+The options for each compiler are set independently. That means that if the GCC-based toolchain is used, but the only
```
boegel: *a* GCC-based toolchain, there are multiple

docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,38 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+compiler specified is ``Intel`` (for example with ``--optarch=Intel:xCORE-AVX2``), then the GCC toolchain will ignore
```
boegel: `, then EasyBuild will behaveas if ``--optarch`` was not specified for this toolchain.`

docs/Controlling_compiler_optimization_flags.rst
```diff
@@ -132,3 +132,38 @@ Currently, using ``--optarch=GENERIC`` will result in the following target archi
...
+that option and behave as ``--optarch`` was not specified.
+
+The compiler name is determined by the ``COMPILER_FAMILY`` constant in the compiler class. At the moment, these are the
+supported names:
+
+* ``GCC``
+* ``Intel``
+* ``PGI``
+* ``Clang``
+* ``IBMXL``
```
boegel: `ClangGCC` and `xlcxlf` are also valid compilers right now (cfr. the respective toolchains where `COMPILER_FAMILY` is being (re)defined.

As soon as we add support for a new compiler (e.g. Open64 or so), this documentation will get outdated, so I'd reword this to:

```
The compiler name corresponds to the value of the ``COMPILER_FAMILY`` constant of the toolchain.
Two common examples are ``GCC`` and ``Intel``.
```

docs/Controlling_compiler_optimization_flags.rst
damianam: You are right. I tried with ``--optarch=GCC:'mach=x86-64 -mtune=generic'``, and it worked, but as soon as I added another compiler, it didn't. Good catch.

boegel:
Thanks for looking into this @damianam!

PR #306, opened 2017-02-13
Fix typo

koallen:
There is a typo in the page `Installation Alternative`, `$HOM` should be `$HOME`


boegel:
Thanks for the typo fix @koallen!

PR #307, opened 2017-02-20
add documentation on contributing back (REVIEW)

boegel:
Preview at http://boegel-eb.readthedocs.io/en/contributing_back/Contributing.html


docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+          If you don't have a GitHub account yet, visit https://github.com/ to create one.
+
+          Make sure you have registered your SSH public key in your GitHub account,
+          via https://github.com/settings/keys.
+
+Before you can open pull requests, you will need to **fork** the GitHub repository
+you are targeting, and create a local **working copy** of that repository. This only
+needs to be done *once* for every of the EasyBuild repositories.
+
+.. note:: For the purpose of this guide, we will be using the ``easybuild-framework`` repository.
```
damianam: You mention later the `--new-pr` option. As far as I know this only works for easyconfigs. I think it would be confusing for a new contributor to see framework as a reference, having an option mentioned, and then see that the option doesn't work in this case.

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+
+In addition to forking the repository on GitHub, you also need to create
+a local *working copy* of the repository. This is basically a local checkout
+of the repository where you can track the changes you are making.
+
+First, *clone* the repository:
+
+.. code:: shell
+
+  git clone git@github.com:hpcugent/easybuild-framework.git
```
damianam: For using the manual approach, I find this confusing too. If you clone your own fork, you don't care about the original repo and don't have to mess with remotes (at least until your copy is out of sync). Wouldn't it make more sense to keep it to the basics now, and simply follow the structure below? It is already complicated enough without introducing additional remotes

fork -> clone own fork -> check out develop -> create new branch -> modify -> commit -> push -> make PR using the webinterface

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+
+Setting up
+~~~~~~~~~~
+
+.. note:: These instructions assume that you already have a working GitHub account.
+
+          If you don't have a GitHub account yet, visit https://github.com/ to create one.
+
+          Make sure you have registered your SSH public key in your GitHub account,
+          via https://github.com/settings/keys.
```
damianam: While I agree that this is highly convenient, I think it doesn't belong here. It is not really a requirement, and it might make users wonder why they have to keep introducing their password even with their key uploaded (that happens easily since the default url uses `https` instead of `git+ssh`)

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+To 'save' your changes, you should create one or more *commits* in the branch
+you created. It is recommended to make separate commits for different 'units of work'.
+
+First, make sure you have checked out the branch were you want to commit the changes to.
+
+For example, to commit changes to the ``mybranch`` branch:
+
+.. code:: shell
+
+  git checkout mybranch
```
damianam: `git commit -m "some message"`

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
++++++++++++++++++++
+
+Once you have committed your changes to a branch, you should *push* your
+branch to your fork of the GitHub repository using ``git push``.
+
+For example, to push the ``mybranch`` branch to the ``myfork`` remote GitHub repository:
+
+.. code:: shell
+
+  git push myfork mybranch
```
damianam: Related to another comment, if the clone is from your own repo, you'd use `git push origin mybranch`, which is more in sync with other git guides out there. Potentially less confusing, since in that case you don't care about other remotes.

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+++++++++++++++++++++++++
+
+To open a pull request using the branch you pushed,
+you should use the GitHub interface, and follow the steps outlined below.
+
+.. note:: Replace '``EXAMPLE``' with your GitHub account name, and
+          '``easybuild-framework``' with the name of the target EasyBuild repository.
+
+i. visit https://github.com/EXAMPLE/easybuild-framework;
+ii. switch to the branch that includes the changes you want to contribute back
```
damianam: Mention explicitly `mybranch`. All the buttons/options are highlighted, except this one.

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+Next, your pull request will be reviewed & tested, see :ref:`contributing_back_review_process`.
+
+.. _contributing_back_updating_pull_requests:
+
+Updating existing pull requests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note:: This section describes the manual procedure to create a new pull request.
+
+          Please consider using ``eb --update-pr`` instead, see :ref:`github_update_pr`.
```
damianam: Mention that this is just for the easyconfigs repo?

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+With this in place, you are all set to open pull requests for your contributions.
+
+.. _contributing_back_creating_pull_requests:
+
+Opening a new pull request
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note:: This section describes the manual procedure to open a new pull request.
+
+          Please consider using ``eb --new-pr`` instead, see :ref:`github_new_pr`.
```
damianam: Mention that this is just for the easyconfigs repo?

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+that was used for opening the pull request, and pushing the updated branch to GitHub.
+
+For example, to update the pull request that was created using the ``mybranch`` branch:
+
+.. code:: shell
+
+  git checkout mybranch
+  # make changes...
+  git commit -am "example commit message for additional changes"
+  git push myfork mybranch
```
damianam: `git push` is enough in this case. Simplifies things, but might create confusion as to why the first time it wasn't necessary. Up to you if you want to simplify the typing

docs/Contributing-back.rst
```diff
@@ -0,0 +1,496 @@
...
+One exception to this rule is *bug fixes*, where the whole point is usually
+to fix functionality that was implemented incorrectly.
+
+This also applies to existing easyconfig files; for example, the versions
+of dependencies should *not* be altered. Adding dependencies that where missing
+or otherwise enhancing existing easyconfigs (e.g., adding extra extensions,
+enabling additional features, etc.) are usually considered acceptable.
+
+In any case, changes that break backward compatibility have to be motivated
+well with technical arguments, and must be approved by the EasyBuild maintainers.
```
damianam: `$DEITY`

boegel:
I will tackle @damianam's remarks very soon (tomorrow morning CET), but I was hoping to get some more feedback on this ASAP.

@migueldiascosta, @bartoldeman, @mboisson, @rjeschmi, @fgeorgatos, @pescobar: Are you up for taking a detailed look at this?
Preferably by reading the rendered preview at http://boegel-eb.readthedocs.io/en/contributing_back/Contributing-back.html, and provide general feedback + inline comments in this pull request.

Please don't hold back from providing feedback even if this was already reviewed by others.
This is likely going to be something that will need some tweaking left and right over time, but I would like to make the first version that gets merged into the official documentation at http://easybuild.readthedocs.io/en/latest/ a good one.

Of particular interest should be the `Review process for contributions` section, since one of my motivations for doing this writeup is to try and get more help in reviewing/testing (easyconfigs) pull requests...

docs/Contributing-back.rst
migueldiascosta: the wiki (https://github.com/hpcugent/easybuild/wiki/Contributing-back) does seem to agree with @damianam, even if we do add a remote and keep it updated (which is what I do), is there a reason to change?

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+you should use the GitHub interface, and follow the steps outlined below.
+
+.. note:: Replace '``EXAMPLE``' with your GitHub account name, and
+          '``easybuild-framework``' with the name of the target EasyBuild repository.
+
+i. visit https://github.com/EXAMPLE/easybuild-framework;
+ii. switch to the branch that includes the changes you want to contribute back
+    using the '``Branch: master``' button on the left;
+iii. click the '``New pull request``' button;
+iv. change the target branch to ``develop`` using the '``base: master``' button;
```
migueldiascosta: I would add "v. review your changes"

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+contributed easyconfig files are from the existing easyconfigs, although it
+does require a bit of practice because of the density of the provided information.
+
+
+.. _contributing_back_review_process_test_reports:
+
+Test reports for easyconfig contributions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For easyconfig contributions, an accompanying **test report must be submitted** to confirm that the touched easyconfig files (still) work as expected.
```
migueldiascosta: by the contributor or the reviewer?

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+is that they **must be accompanied by additional tests or test cases**.
+
+For new features or enhancements, a dedicated test (case) must be added
+which verifies that the feature implementation works as expected.
+
+For bug fixes, a test (case) must be added that triggers the code path where the bug manifested,
+and which verifies that the bug was indeed fixed.
+
+Tests not only confirm that the implementation is correct, it also helps to ensure
+that any future changes will not affect semantics.
```
rjeschmi: I don't think semantics is the right word here.

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+
+.. _contributing_back_review_process_backward_compatibility:
+
+Backward compatibility
+~~~~~~~~~~~~~~~~~~~~~~
+
+**Contributions should retain backward compatibility**, i.e., they should *not*
+make any changes that alter the (default) semantics of the existing code base.
```
rjeschmi: I don't think semantics is the right word here either. I think interface is probably better. Semantics implies a much more rigorous interpretation of the code in its context. I think backwards compatibility is actually clearer here than the explanation.

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+Of course, enhancements to existing code that retain backward compatibility can be made.
+
+One exception to this rule is *bug fixes*, where the whole point is usually
+to fix functionality that was implemented incorrectly.
+
+This also applies to existing easyconfig files; for example, the versions
+of dependencies should *not* be altered. Adding dependencies that where missing
```
rjeschmi: s/where/were/

docs/Contributing-back.rst
rjeschmi: "in any case" should be removed anyway. It is too informal considering the previous context (pretty much a voice change).

docs/Contributing-back.rst
boegel: OK, I'll change this to a recommendation.

Note that the instructions specify to clone via `git@`, so no `https://` is being used if you follow these instructions.

docs/Contributing-back.rst
boegel: The intention is that `--new-pr` also supports framework & easyblocks soonish, so I'm trying to avoid to having to update this page accordingly...

The `--new-pr` documentation where we refer to makes it clear already that it, for now, only works with easyconfigs.

docs/Contributing-back.rst
boegel: You need upstream to be able to keep your `develop` branch in sync (which is actually missing here), which is why I added it.

But you are right that cloning your own fork may make more sense, I'll change that, and add a dedicated subsection explaining how to keep the ``develop`` branch in your working directory in sync with upstream.

Thanks for the suggestion y'all!

docs/Contributing-back.rst
boegel: I'll do this in the `--new-pr` documentation itself, since it's less likely we'll forget to update that once ``easybuild-framework`` and ``easybuild-easyblocks`` are also supported in ``--new-pr`` and ``--update-pr``

docs/Contributing-back.rst
boegel: no, this is the preparation for committing to the `mybranch` branch; I'll clarify that

docs/Contributing-back.rst
boegel: ok, changed

docs/Contributing-back.rst
boegel: ok

docs/Contributing-back.rst
boegel: done in a note in the `--update-pr` documentation

docs/Contributing-back.rst
boegel: What is being pushed using just `git push` depends on your Git config, so I'd rather be explicit here

docs/Contributing-back.rst
boegel: ok

docs/Contributing-back.rst
```diff
@@ -15,39 +15,153 @@ and discusses the review process for contributions.
...
+requires being intimately familiar with its codebase.
+
+There are various ways to contribute back, even if you don't have any experience
+with the programming language(s) being used by the project you are interested in.
+
+You can contribute back to EasyBuild by:
+
+* :ref:`contributing_back_feedback`
+* :ref:`contributing_back_bugs`
+* :ref:`contributing_back_request`
```
boegel: fixed, thanks

docs/Contributing-back.rst
```diff
@@ -15,39 +15,153 @@ and discusses the review process for contributions.
...
+* :ref:`contributing_back_helping_others`
+* :ref:`contributing_back_easyconfigs`
+* :ref:`contributing_back_code`
```
boegel: ok, good catch, thanks

docs/Contributing-back.rst
boegel: makes sense, thanks

docs/Contributing-back.rst
boegel: preferably by the contributor, but this can also be done by a 3rd party

I'll clarify that, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
+.. _contributing_back:
+
+Contributing back
```
boegel: ok, I'll change that, thanks for the suggestion

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+.. _contributing_back_feedback:
+
+Giving feedback
+~~~~~~~~~~~~~~~
+
+An easy way to contribute back to EasyBuild, even without having a lot of
+hands-on experience with it, is to **provide feedback** on your experiences.
+
+Feedback from people new to EasyBuild is very valuable, since it provides
```
boegel: yeah, I have a tendency to be a bit wordy... keep suggestions like this coming!

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+A particularly useful tool is http://rst.ninjs.org/, which can be used to
+preview how the documentation written in ``reStructuredText`` format
+will be rendered on ``readthedocs.org`` (select the ``Nature`` theme for optimal
+results).
+
+
+.. _contributing_back_discuss:
+
+Joining the discussion
```
boegel: ok

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+Last but not least, you can actively join the discussions that arise on the EasyBuild
+mailing list, the ``#easybuild`` IRC channel and during the bi-weekly EasyBuild
+conference calls (see https://github.com/hpcugent/easybuild/wiki/Conference-calls).
+
+Engage with the EasyBuild community, and help steer EasyBuild development by
+participating in the often lively (yet civil and respectful) discussions regarding
```
boegel: replaced `lively (yet civil and respectful) discussions` with `conversations` (which is a bit more boring, but fine ;))

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+          Make sure you have registered your SSH public key in your GitHub account,
+          via https://github.com/settings/keys.
+
+Before you can open pull requests, you will need to **fork** the GitHub repository
+you are targeting, and create a local **working copy** of that repository. This only
+needs to be done *once* for every of the EasyBuild repositories.
+
+.. note:: For the purpose of this guide, we will be using the ``easybuild-framework`` repository.
+
+          However, the instructions are analogous for the other EasyBuild repositories.
```
boegel: ok, `equivalent` it is, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+.. code:: shell
+
+  git add easybuild/tools/filetools.py
+  git status  # check which files are staged for commit
+  git commit -m "example commit message for changes to filetools.py"
+
+.. note:: Please use a concise commit message that describes the changes you made.
+
+To create a single commit for *all* the changes that were made, use ``git commit -am "<message"``.
```
boegel: ok, I overlooked the fact that files need to be tracked, I'll just drop this then

docs/Contributing-back.rst
boegel: ok, fixed

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+* test reports must be submitted for easyconfig contributions (see :ref:`contributing_back_review_process_test_reports`)
+
+
+.. _contributing_back_review_process_travis:
+
+Test suite (Travis)
+~~~~~~~~~~~~~~~~~~~
+
+Each pull request is tested automatically by Travis, and the test result
```
boegel: ok, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+.. _contributing_back_review_process:
+
+Review process for contributions
+--------------------------------
+
+Each contribution is thoroughly reviewed and tested before it gets merged in.
+Some aspects of this are fully automated, others require human intervention.
+
+It is important to be aware of all of the aspects of the review process,
```
boegel: thanks, that's better indeed

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+Usually, merging a pull request implies that the changes will be part of the next EasyBuild release.
+
+
+.. _contributing_back_review_process:
+
+Review process for contributions
+--------------------------------
+
+Each contribution is thoroughly reviewed and tested before it gets merged in.
+Some aspects of this are fully automated, others require human intervention.
```
boegel: ok

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+Merging of pull requests
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once your pull request have been given the green light by Travis,
+and one or more people reviewing have approved the changes, it can be merged into
+the ``develop`` branch. This can only be done by someone in the EasyBuild
+'*admin*' team.
+
+Usually, merging a pull request implies that the changes will be part of the next EasyBuild release.
```
boegel: ok, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+.. _contributing_back_merging_pull_requests:
+
+Merging of pull requests
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once your pull request have been given the green light by Travis,
+and one or more people reviewing have approved the changes, it can be merged into
+the ``develop`` branch. This can only be done by someone in the EasyBuild
+'*admin*' team.
```
boegel: ok, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+Updating a pull request will trigger Travis to re-test your contribution,
+and a notification will be sent out to whoever is 'watching' your pull request.
+
+
+.. _contributing_back_merging_pull_requests:
+
+Merging of pull requests
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once your pull request have been given the green light by Travis,
```
boegel: ok, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+
+Updating existing pull requests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note:: This section describes the manual procedure to create a new pull request.
+
+          Please consider using ``eb --update-pr`` instead, see :ref:`github_update_pr`.
+
+It is quite common to update a pull request after creating it, for example
+because Travis reports problems with the changes being made, or as a response
```
boegel: ok, thanks

docs/Contributing-back.rst
```diff
@@ -0,0 +1,610 @@
...
+Once you have committed your changes to a branch, you should *push* your
+branch to your fork of the GitHub repository using ``git push``.
+
+For example, to push the ``mybranch`` branch to the ``myfork`` remote GitHub repository:
+
+.. code:: shell
+
+  git push myfork mybranch
+
+Note that this is equivalent to making your work public.
```
boegel: ok,thanks

docs/Contributing-back.rst
boegel: used `functionality` instead

docs/Contributing-back.rst
boegel: oops, thanks

boegel:
Going in, thanks a lot for the feedback @damianam, @rjeschmi and @migueldiascosta!

PR #330, opened 2017-04-13
Correct location for example config file

pol:
The path was not the default configuration file location (though maybe it was in the past).


boegel:
Nice catch @pol!

This was indeed changed a long time ago (EasyBuild v1.11.0, released Feb 16th 2014...).

pol:
No problem! I am happy to help.

PR #365, opened 2017-07-20
included --backup-module documentation

damianam:
This is the documentation for https://github.com/easybuilders/easybuild-framework/pull/2134

PR #367, opened 2017-08-24
document --backup-modules

boegel:
Support for `--backup-modules` was merged to be included in EasyBuild v3.4.0 in https://github.com/easybuilders/easybuild-framework/pull/2134.

This leverages #365; since @damianam is on leave I looked into extending it a bit.

Preview at http://boegel-eb.readthedocs.io/en/backup_modules/Backup_modules.html

### DIFF
--- a/docs/Backup_modules.rst
+++ b/docs/Backup_modules.rst
@@ -24,8 +24,8 @@ The location of the backed up module file will be printed, as well as a "unified
 
 .. _backup_modules_disable:
 
-Disabling backup of modules
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Disabling automatic backup of modules
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When ``--skip`` or ``--module-only`` is used, backing up of existing modules is enabled automatically.
 
--- a/docs/Backup_modules.rst
+++ b/docs/Backup_modules.rst
@@ -1,0 +1,96 @@
+.. _backup_modules:
+
+Backing up of existing modules (``--backup-modules``)
+-----------------------------------------------------
+
+While regenerating existing module files, you may want to preserve the existing module files
+to compare and assess that the changes in the newly generated module file match expectations.
+
+Backing up of existing modules can be enabled with ``--backup-modules``.
+
+Backups are stored in the same directory as where the module file was located,
+and the files are given an additional extention of the form ``.bak_<year><month><day><hour><min><sec>``.
+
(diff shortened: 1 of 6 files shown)


docs/Backup_modules.rst
```diff
@@ -0,0 +1,96 @@
...
+
+This can be disabled with ``--disable-backup-modules``.
+
+
+.. _backup_modules_example:
+
+Example
+~~~~~~~
+
+Suppose existing modules in both Tcl & Lua syntax are present for bzip2 1.0.6.
```
ocaisa: Change to be similar to what is below:
```
 ``bzip2/1.0.6``
```

docs/Backup_modules.rst
```diff
@@ -0,0 +1,96 @@
...
+
+    $ ls -la $EASYBUILD_PREFIX/modules/all/bzip2
+    total 32
+    drwxr-xr-x  2 example  example   204 Aug 24 10:26 .
+    drwxr-xr-x  3 example  example   102 Jul 11 10:18 ..
+    -rw-r--r--  1 example  example  1227 Aug 24 10:24 .1.0.6.bak_20170824102412
+    -rw-r--r--  1 example  example  1256 Jul 11 01:24 1.0.6
+    -rw-r--r--  1 example  example  1303 Jul 11 01:26 1.0.6.lua
+    -rw-r--r--  1 example  example  1259 Aug 24 10:26 1.0.6.lua.bak_20170824102603
+
```
ocaisa: I'm curious what happens in the case that you have an initial module in tcl syntax and then generate the replacement in lua syntax (or vice versa). Can the option handle that?

boegel: If you have a module in Tcl syntax, and then generate a module in Lua, the Tcl one is untouched, and hence not backed up.

Backing up is only relevant when you're actually overwriting an existing file.

ocaisa: Hmm, but isn't that exactly a use case here...I have tcl modules and then I move to Lmod, I want to update to Lua syntax...but if I leave the tcl module untouched it will automatically be picked up by Lmod, so I have 2 available modules providing the same software package (or does Lmod always choose lua files first?)

ocaisa: I guess I should have asked that in the original PR though, this PR is good to go

boegel: Lmod always prefers Lua modules over Tcl modules; if you have both, the Tcl module is ignored.

With Tmod, `.lua` modules are not picked up because the `#%Module` header isn't there.

ocaisa: Ok, that works for me

PR #392, opened 2017-12-08
document support for user-defined hooks (WIP)

boegel:
documentation update for feature implemented in https://github.com/easybuilders/easybuild-framework/pull/2343

preview at http://boegel-eb.readthedocs.io/en/hooks_docs/Hooks.html

### DIFF
--- a/docs/Hooks.rst
+++ b/docs/Hooks.rst
@@ -1,0 +1,143 @@
+.. _hooks:
+
+Hooks
+=====
+
+Since EasyBuild v3.5.0, support is available for implementing *hooks* that can be used
+to customise the behaviour of EasyBuild according to site policies if needed,
+without having having to change the EasyBuild framework or the existing easyblocks.
+
+.. contents::
+    :depth: 3
+    :backlinks: none
+
+.. _hooks_what:
+
+What are hooks?
+---------------
+
+*Hooks* are user-defined functions that are called by the EasyBuild framework at specific times during
+the installation procedure that is performed for each software package being installed.
+
+They can be leveraged to alter or augment the installation procedure in a particular way.
+This is usually done to conform with site-specific policies that are difficult to enforce otherwise,
+but it can also be (ab)used to fix specific problems or inject self-implemented enhancements
+(before you flesh them out in a proper contribution, for example).
+
+Both the ``EasyBlock`` instance and the parsed easyconfig file that are being used
+are fully accessible (and modifiable) from within hook implementations, Hence, this mechanism
+provides a lot of flexibility to change the EasyBuild functionality should you require it,
+without having to modify the codebase of EasyBuild itself.
+
+
+.. _hooks_configuration:
+
+Configuring EasyBuild to use your hook implementations
+------------------------------------------------------
+
+To instruct EasyBuild to use your hook implementations,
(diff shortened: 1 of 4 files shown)


docs/Hooks.rst
```diff
@@ -0,0 +1,143 @@
+.. _hooks:
+
+Hooks
+=====
+
+Since EasyBuild v3.5.0, support is available for implementing *hooks* that can be used
+to customise the behaviour of EasyBuild according to site policies if needed,
+without having having to change the EasyBuild framework or the existing easyblocks.
```
damianam: having having?

docs/Hooks.rst
```diff
@@ -0,0 +1,143 @@
...
+*Hooks* are user-defined functions that are called by the EasyBuild framework at specific times during
+the installation procedure that is performed for each software package being installed.
+
+They can be leveraged to alter or augment the installation procedure in a particular way.
+This is usually done to conform with site-specific policies that are difficult to enforce otherwise,
+but it can also be (ab)used to fix specific problems or inject self-implemented enhancements
+(before you flesh them out in a proper contribution, for example).
+
+Both the ``EasyBlock`` instance and the parsed easyconfig file that are being used
+are fully accessible (and modifiable) from within hook implementations, Hence, this mechanism
```
damianam: `implementations. Hence`

docs/Hooks.rst
```diff
@@ -0,0 +1,143 @@
...
+
+
+.. _hooks_available:
+
+Available hooks
+---------------
+
+Currently, two types of hooks are supported:
+
+* ``start_hook`` and ``end_hook``, which are triggered before starting a new software installation, and right after,
```
damianam: I thought those hooks were called at the beginning and end of the installation "session" so to say. Not per package. I guess I was wrong, but probably that makes sense. In our case, for instance, we would like to refresh the cache at the end of a "session", not every time a new dependency is installed.

docs/Hooks.rst
```diff
@@ -0,0 +1,143 @@
...
+  respectfully;
+* hooks that a triggered before and after every step of the installation procedure, aptly named
```
damianam: hooks that are triggered

docs/Hooks.rst
boegel: I can think of use cases for both, so we should add support for an additional set of hooks `start_session_hook` and `end_session_hook`?

That may call for renaming the existing `start_hook` and `end_hook` to something less ambiguous, like `start_easyconfig_hook` and `end_easyconfig_hook`?

I'm open to suggestions for better names...

damianam: `init_hook` and `finalize_hook`? And `pre_easyconfig_hook`/`post_easyconfig_hook`?

docs/Hooks.rst
```diff
@@ -0,0 +1,152 @@
...
+
+.. _hooks_what:
+
+What are hooks?
+---------------
+
+*Hooks* are user-defined functions that are called by the EasyBuild framework at specific times during
+the installation procedure that is performed for each software package being installed.
+
+They can be leveraged to alter or augment the installation procedure in a particular way.
```
damianam: in any way?

damianam:
Let's merge this, lgtm

PR #408, opened 2018-01-31
Initial version of GCCcore description

verdurin:
First draft of a description of the difference between `GCC` and `GCCcore`, based on the comment in https://github.com/easybuilders/easybuild-easyconfigs/pull/5579

### DIFF
--- a/docs/Common-toolchains.rst
+++ b/docs/Common-toolchains.rst
@@ -87,6 +87,34 @@ The ``intel`` common compiler toolchain consists of the Intel compilers and libr
           valid licenses must be available to install and use it.
 
 
+.. _common_toolchains_gcccore:
+
+``GCCcore`` and ``GCC`` common toolchains
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+There are two lower-level toolchains including the ``GCC`` compiler.
+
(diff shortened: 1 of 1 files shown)


docs/Common-toolchains.rst
```diff
@@ -87,6 +87,34 @@ The ``intel`` common compiler toolchain consists of the Intel compilers and libr
...
+installed with the ``dummy`` toolchain (i.e. which were built with the system
+compiler). One level up, you have tools that are built with a compiler-only
+toolchain (and higher levels are built with a compiler+MPI toolchain, or a
+full toolchain, i.e. compiler+MPI+BLAS/LAPACK/FFTW).
+
+For more information on a setup like this, see http://easybuilders.github.io/easybuild/files/hust14_paper.pdf .
+
+Easyconfigs that use the ``GCCcore`` toolchain result in modules on the
+compiler-only level of a hierarchy, which means those modules can be used to
+resolve dependencies for easyconfigs that use several toolchains, i.e. both the
```
akesandgren: Perhaps something like this instead:
"resolve dependencies for easyconfigs that use either of the"
or something similar. The current text is a bit odd. A single easyconfig can only use one toolchain, but the above suggests that it can use multiple toolchains.

PR #412, opened 2018-02-13
add policy on closing prs

migueldiascosta:


boegel:
Looks great to me!

Thoughts on this @easybuilders/easybuild-easyconfigs-maintainers?

ocaisa:
Sounds reasonable to me, we can update it if necessary after the first round of implementation

PR #425, opened 2018-04-25
Improved description of behavior with --allow-loaded-modules

vanzod:
Improved the clarity of the EasyBuild behavior when modules are listed as allowed via `--allow-loaded-modules`.


boegel:
Excellent, thanks a lot @vanzod!

PR #432, opened 2018-05-09
remark that git-working-dirs-path option can be set in the config file

arielzn:
I propose to add this minor remark in the github integration docs, it might be obvious but it is not explicitly mentioned in the docs.

### DIFF
--- a/docs/Integration_with_GitHub.rst
+++ b/docs/Integration_with_GitHub.rst
@@ -78,7 +78,7 @@ To specify your GitHub username, do one of the following:
 
 * use the ``--github-user`` configuration option on the ``eb`` command line
 * define the ``$EASYBUILD_GITHUB_USER`` environment variable
-* specify ``github-user`` in your EasyBuild configuraton file
+* specify ``github-user`` in your EasyBuild configuration file
 
 (see also :ref:`configuring_easybuild`)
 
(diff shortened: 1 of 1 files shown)


docs/Integration_with_GitHub.rst
```diff
@@ -130,7 +130,8 @@ EasyBuild will validate the provided token, to check that authenticated access t
 Specify location of working directories (``--git-working-dirs-path``)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can specify the location of your Git working directories using ``--git-working-dirs-path``.
+You can specify the location of your Git working directories using ``--git-working-dirs-path`` or
+you can set it with the ``git-working-dirs-path`` option in your EasyBuild configuraton file.
```
boegel: There actually a 3rd option (just like for any other configuration setting): specifying the location via the `$EASYBUILD_GIT_WORKING_DIRS_PATH` environment variable, see http://easybuild.readthedocs.io/en/latest/Configuration.html#consistentency-across-supported-configuration-types...

So either we mention all three options, or maybe just refer to that section?

Rewording `using --git-working-dirs-path` to `using the git-working-dirs-path configuration setting` probably also helps?

arielzn: ok, so maybe it can be mentioned the same as for `Providing a GitHub username` ?
with three bullets items?

boegel: Sure, that makes sense.

On the other hand, if we need to do this for every configuration setting we mention somewhere, that may be a bit too much...

Feel free to change it in this specific case those, I'll be happy to merge it!

boegel:
lgtm, thanks a lot @arielzn! Should show up at http://easybuild.readthedocs.io/en/latest/Integration_with_GitHub.html#specify-location-of-working-directories-git-working-dirs-path in a couple of minutes...

PR #443, opened 2018-07-10
add definitions of 2018b common toolchains

boegel:


bartoldeman:
lgtm, double checked the version numbers.

PR #468, opened 2018-11-07
add documentation on deprecated easyconfigs/toolchains

boegel:
see also https://github.com/easybuilders/easybuild-framework/pull/2656

### DIFF
--- a/docs/Deprecated-easyconfigs.rst
+++ b/docs/Deprecated-easyconfigs.rst
@@ -15,14 +15,18 @@ Since EasyBuild v3.8.0, individual easyconfig files or particular (versions of) 
 Symptoms
 --------
 
-Using an easyconfig file or toolchain that was marked as deprecated results in a warning message like::
+Using an easyconfig file or toolchain that was marked as deprecated results in a warning message like:
+
+.. code:: text
 
   WARNING: Deprecated functionality, will no longer work in v4.0: easyconfig file '/home/example/test.eb' is marked as deprecated:
   This is an example message explaining why the easyconfig file was deprecated.
   (see also http://easybuild.readthedocs.org/en/latest/Deprecated-easyconfigs.html)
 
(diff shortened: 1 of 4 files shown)


boegel:
@ocaisa Wasn't there yet because there was an issue with readthedocs, but it's there now, see https://boegel-eb.readthedocs.io/en/deprecated_ecs_tcs/Deprecated-easyconfigs.html

docs/Deprecated-easyconfigs.rst
```diff
@@ -0,0 +1,53 @@
...
+
+
+.. _deprecated_easyconfigs_symptoms:
+
+Symptoms
+--------
+
+Using an easyconfig file or toolchain that was marked as deprecated results in a warning message like::
+
+  WARNING: Deprecated functionality, will no longer work in v4.0: easyconfig file '/home/example/test.eb' is marked as deprecated:
```
ocaisa: These blocks are being syntax highlighted (for Python I think). Not a sphinx expert but should be easy to switch off.

boegel: You would expect this to be easy to fix, turns out it's not...

With `.. code:: none`, the block is not displayed at all, with `.. code:: text` I'm still getting highlighting.

I'll keep poking at it, thanks for pointing this out, it indeed looks a bit silly this way...

ocaisa:
Seems to be ok now, if you're ok with it I can merge

ocaisa:
According to https://stackoverflow.com/questions/44376893/selectively-disable-readthedocs-syntax-highlighting you also could have used
``
.. highlight:: none
``

boegel:
@ocaisa Tried `.. highlight:: none` via http://rst.ninjs.org/, and it barfed on it...

Anyway, problem fixed indeed, good to go imho!

ocaisa:
Yeah, I think it's a sphinx option as opposed to an rst option

PR #469, opened 2018-11-18
Update Configuration.rst

andreas-h:
make path to template configuration file consistent


boegel:
Thanks for the fix @andreas-h!

PR #484, opened 2019-02-19
Run codespell on docs

ocaisa:
Use the codespell python package to spell check the docs, see https://github.com/codespell-project/codespell for details.

I didn't clean `Supported_software.rst` since that is dynamically generated (I think).

### DIFF
--- a/docs/version-specific/generic_easyblocks.rst
+++ b/docs/version-specific/generic_easyblocks.rst
@@ -233,7 +233,7 @@ Build a Python package and module with cmake.
     Some packages use cmake to first build and install C Python packages
     and then put the Python package in lib/pythonX.Y/site-packages.
 
-    We install this in a seperate location and generate a module file 
+    We install this in a separate location and generate a module file 
     which sets the PYTHONPATH.
 
     We use the default CMake implementation, and use make_module_extra from PythonPackage.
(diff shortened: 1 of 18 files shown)


docs/version-specific/generic_easyblocks.rst
```diff
@@ -233,7 +233,7 @@ Build a Python package and module with cmake.
     Some packages use cmake to first build and install C Python packages
     and then put the Python package in lib/pythonX.Y/site-packages.
 
-    We install this in a seperate location and generate a module file 
+    We install this in a separate location and generate a module file
```
boegel: @ocaisa This is a typo in a docstring of the generic `CMakePythonPackage` easyblock, should also be fixed there or it'll be re-introduced (all docs under `docs/version-specific` are re-generated automatically with every EasyBuild release)

boegel:
Thanks a lot @ocaisa!

Can we come up with a way of making Travis fail whenever spelling errors like this are introduced in the documentation?

ocaisa:
I think this should be possible, not a travis expert though but I can give it a try

ocaisa:
We just need a list of directories that should be excluded, like `.git` and the automatically generated directories

ocaisa:
@boegel Ok I have two tests that you could use as CI on this repo, first for a spell check, second for a build check:
```
codespell --skip=".git,version-specific" --ignore-words-list=atleast --quiet-level=2; retVal=$?; if [ $retVal -eq 1 ]; then echo "Got one expected warning, so success!";  else $(exit $retVal); fi;
READTHEDOCS=1 sphinx-build docs test
```
These depend on the `sphinx` and `codespell` python packages

PR #485, opened 2019-02-21
Add CI to docs

ocaisa:


ocaisa:
First attempt at adding CI, open to suggestions @boegel as per #484

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: python
+python:
+  - "2.7"
+  - "3.6"
+# command to install dependencies
+install:
+  - pip install sphinx codespell
+# command to run tests
+script:
+  - codespell --skip=".git,version-specific" --ignore-words-list=atleast --quiet-level=2; retVal=$?; if [ $retVal -eq 1 ]; then echo "Got one expected warning, so success!";  else $(exit $retVal); fi;
```
boegel: What is the expected warning?

What if `retVal=0`, then we'll do an `exit 0`?

.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: python
+python:
+  - "2.7"
+  - "3.6"
```
boegel: There's no point in testing with two Python versions, since Python is only required for the tools we're using, we're not testing any Python code here?

I'd stick to Python 3.6 if that works, more future proof...

docs/conf.py
```diff
@@ -41,7 +41,7 @@
 # other places throughout the built documents.
 #
 # The short X.Y version.
-version = '3.9.0dev'  # this is meant to reference the version of EasyBuild
+version = '3.8.1'  # this is meant to reference the version of EasyBuild
```
boegel: Please reset this to `3.9.0dev`, since we're in `develop` here...

.travis.yml
ocaisa: The gymnastics here are because of the disabled suggestion because of the use of a proper name 
```
./docs/Cray-support.rst:19: Tim  ==> Time, tim  | disabled due to being a person's name
```
(see https://travis-ci.org/ocaisa/easybuild/jobs/496488317#L493). There was no way to turn this off (that I could find) and the return value is the number of messages it prints, so if this one is expected then we always see at least one, if it's 0 then even better.

boegel: Maybe add a comment to clarify that?

boegel:
close/open to trigger Travis (now it's enabled for this repo)

boegel:
lgtm, thanks @ocaisa!

PR #487, opened 2019-02-22
add Samuel Moors as EasyBuild maintainer

boegel:
Let's give @smoors the honor of merging this himself... :)


smoors:
looks very good to me :)

PR #503, opened 2019-04-26
add documentation on writing easyconfigs for Python & (bundles) of Python packages (WIP)

boegel:
@micketeer @akesandgren Suggestions for topics that should be covered but are not listed yet are welcome...

I plan to work on this section by section, and hopefully finishing up on it soon, since we need this dearly...

preview @ https://boegel-eb.readthedocs.io/en/python/software-specific/Python.html


surak:
How about docu for the pypi options? It's not easy to find

PR #513, opened 2019-05-25
update docs on container integration w.r.t. enhanced --containers-base-config

boegel:
updated docs for changes made in https://github.com/easybuilders/easybuild-framework/pull/2903

rendered preview at https://boegel-eb.readthedocs.io/en/containers/Containers.html

### DIFF
--- a/docs/Containers.rst
+++ b/docs/Containers.rst
@@ -91,6 +91,9 @@ There are a couple of specific requirements for the base container image:
 
 The ``easybuild`` user will be used when running EasyBuild to install the specified software stack.
 
+An example recipe for a Singularity container image that can be used as a base image is avaiilable at
+https://singularity-hub.org/containers/1569/download/recipe .
+
 .. note:: The generated container recipe currently hardcodes some of this.
           We intend to make this more configurable in a future version of EasyBuild.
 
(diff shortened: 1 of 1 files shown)


docs/Containers.rst
```diff
@@ -57,31 +57,42 @@ that corresponds to the easyconfig files that are specified as arguments to the
...
+
+Requirements for base container image
++++++++++++++++++++++++++++++++++++++
+
+There are a couple of specific requirements for the base container image:
+
+* all dependencies of EasyBuild must be installed, incl. Lmod (cfr. :ref:`requirements`)
+* a user named ``easybuild`` must be available
+* the ``/scratch`` and ``/app`` directories must exist,
+  and the ``easybuild`` user must have write permissions to those directories
```
wpoely86: do we have an example bootstrap file to create such an image to which we can point?

boegel: I could include a link to https://singularity-hub.org/containers/1569/download/recipe (not sure how long that'll stay around), but the idea is to actually remove these hardcoded requirements in the near term, and i) have support in `eb` to generate a base container from scratch, ii) allow tweaking each of these (username + installation prefix + location for temporary stuff).

wpoely86: I would add the link. It's much easier to give a working recipe then to tell what should be done.

boegel: I'm already using that container in the section on `--container-base-image`, and I'm not a big fan of relying heavily on it.

If things work as planned, we can get rid of these requirements by providing additional keywords in `--container-base-config` for these, so they can be tweaked as needed.

In fact, when you already supply your own template recipe, these are not relevant, since you can control things yourself that way already.

PR #514, opened 2019-06-05
update documentation on Singuluarity integration w.r.t. recent changes

boegel:
Covers changes made in:
* https://github.com/easybuilders/easybuild-framework/pull/2907
* https://github.com/easybuilders/easybuild-framework/pull/2909
* https://github.com/easybuilders/easybuild-framework/pull/2910

preview at https://boegel-eb.readthedocs.io/en/containers/Containers.html

### DIFF
--- a/docs/Containers.rst
+++ b/docs/Containers.rst
@@ -18,6 +18,8 @@ container software of choice to create *container images*.
 Initially (since EasyBuild v3.6.0), only Singularity (https://www.sylabs.io/singularity) was supported.
 Since EasyBuild v3.6.2, generating (recipes for) Docker (https://www.docker.com/) containers is also supported.
 
+In EasyBuild v3.9.2 the support for generating Singularity container recipes/images was enhanced significantly.
+
 .. contents::
     :depth: 3
     :backlinks: none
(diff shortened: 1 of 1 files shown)


docs/Containers.rst
```diff
@@ -331,28 +316,63 @@ Example of a generated container recipe
...
-    # upgrade easybuild package automatically to latest version
-    pip install -U easybuild
+    yum install --quiet --assumeyes epel-release
+    yum install --quiet --assumeyes python setuptools Lmod
+    yum install --quiet --assumeyes bzip2 gzip tar unzip xz
+    yum install --quiet --assumeyes curl wget
+    yum install --quiet --assumeyes patch make
+    yum install --quiet --assumeyes file git which
+    yum install --quiet --assumeyes python-pip
+    yum install --quiet --assumeyes gcc-c++
```
wpoely86: why are this all separate calls to yum? It will be quicker when combined.

docs/Containers.rst
```diff
@@ -196,6 +157,30 @@ is downloaded from the corresponding registry, ad specified through the ``from``
...
+Requirements for base container image
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are a couple of specific requirements for the base container image:
+
+* all dependencies of EasyBuild must be installed, including:
+
+  * Python 2.6 or 2.7
+  * Lmod
+  * standard tools & utilities like ``make``, ``patch``, ``tar``, etc.
```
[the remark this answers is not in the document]
boegel: No, some things will be missing (e.g. `bzip2` is not there by default, it seems).

Making the list complete here will be a PITA to maintain, I'd rather maintain it only in the code.

In practice, people should start with a base image created via a recipe generated by EasyBuild, then all should be fine.

docs/Containers.rst
boegel: That's just because they're listed separately in the code, I doubt it'll make a huge difference when we do everything in one go though?

boegel: Just timed it, total time (incl. `pip install easybuild` and `eb bzip2-1.0.6.eb`), I don't see much (if any) difference between running multiple `yum install` commands vs one big `yum install`.

In addition, the `yum install epel-release` has to be done separately, otherwise installing `python-pip` and `Lmod` fails.

I don't think it's worth making the generated recipe harder to read for humans for a matter of seconds at best?

PR #515, opened 2019-06-07
final docs update w.r.t. container integration for EasyBuild v3.9.2

boegel:
preview at https://boegel-eb.readthedocs.io/en/containers_update/Containers.html


docs/Containers.rst
```diff
@@ -80,14 +80,17 @@ When generating container recipes, EasyBuild will replace the following template
   * this is expected to include lines that specify ``From:``, ``MirrorURL:``, etc.
   * for more information, see :ref:`containers_usage_config`
 
+* ``%(easyconfigs)s``: (list of) easyconfig file name(s)/path(s) to pass to ``eb`` command
+* ``%(eb_args)s``: additional arguments for 'eb' command
+* ``%(include)s``: list of additional OS packages to include
```
ocaisa: The usage of this is not really clear.  It makes it sound like I can list openssl and friends here rather than use `install_os_deps`, is that right?

boegel: Yes and no, this is really intended for very basic stuff like `yum`, see https://easybuild.readthedocs.io/en/develop/Containers.html#include-keyword-os-packages-to-include

ocaisa: Yeah, I gathered that from the example but wanted to be sure, probably could be clarified but it's a separate issue to this PR

ocaisa: Well, you could add ` (e.g., 'yum')`

boegel: The template values that you can use in a template recipe you provide are of course tied to how those values will be filled out, I guess referring to the section of the `include` keyword from here could help?

ocaisa: Sorry, I wasn't paying enough attention to what was being described here. Yes a link to https://boegel-eb.readthedocs.io/en/containers_update/Containers.html#include-keyword-os-packages-to-include would be good.

PR #557, opened 2019-11-07
also mention EnvironmentModules as possible configuration value for ---modules-tool

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #558, opened 2019-11-12
add subsection on use of wildcards in --external-modules-metadata

boegel:
see also https://github.com/easybuilders/easybuild-framework/pull/3075


migueldiascosta:
Going in, thanks @boegel!

PR #562, opened 2019-11-19
Updated usage of install-EasyBuild-develop.sh

terjekv:
- Specify dependency of a Python installation with GitPython.
  - Document depenceny on vsc-install.


ocaisa:
@boegel The script (and docs linked here) need an update to reflect the changes in v4.x since `vsc-*` are no longer required

PR #565, opened 2019-11-24
first draft governance model (bfdl)

shahzebsiddiqui:
first draft Governance Model

### DIFF
--- a/GOVERNANCE.rst
+++ b/GOVERNANCE.rst
@@ -1,0 +1,134 @@
+.. _maintainers:
+
+Goverance Document
+====================
+
+This document describes the governance model of the Easybuild Project.
+
+Overview
+----------
+
+The Easybuild Project  is an open source software and released under GNU General Public License v2.0 and developed openly 
+and hosted in public GitHub repositories under the `easybuilders <https://github.com/easybuilders>`_ organization. 
+
+The Project is developed by a team of distributed developers, called Contributors. Contributors are individuals who have 
+contributed code, documentation, designs or other work to one or more Project repositories. Anyone can be a Contributor. 
+Contributors can be affiliated with any legal entity or none. Contributors participate in the project by submitting, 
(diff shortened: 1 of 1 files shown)


boegel:
@shahzebsiddiqui Thank you for looking into this!

There's a couple of things I don't like here, like the BDFL part, and the fact that we have maintainers listed in two different places.

This document should just point to https://easybuild.readthedocs.io/en/latest/Maintainers.html imho... The corresponding file is also in this repo btw, see https://github.com/easybuilders/easybuild/blob/develop/docs/Maintainers.rst

shahzebsiddiqui:
@boegel  Yes i understand the file https://github.com/easybuilders/easybuild/blob/develop/docs/Maintainers.rst was part of documentation and I wrote the GOVERNANCE file with assumption that this would reside outside of documentation build. Generally it should be easily found in the top level repo. 

The governance file should address how the easybuild project is governed taking into account the three repos (``easybuild-framework``, ``easybuild-easyblocks``, ``easybuild-easyconfigs``). That way you can have one governance file for the entire project instead of three separate governance file per repo.

Eventually you may want to consider removing the Maintainers.rst from documentation once the Governance File is finalized so dont worry if it is documented in two locations for time being.

It would be worth breaking down the role for ``Easybuild Maintainers`` to sub-committees so roles are defined clearly and new users can address questions to appropriate committee when requesting for support assistance. For instance you may want to have core-maintainers for each repo depending on contribution and experience level. Also there needs to be section on how frequent contributors can transition to core-contributors. It is a gray-line but you may want to decide at some point to give more power to the contributors to be part of the community and maintain the project and maybe require them to attend Easybuild Conf Call on regular basis.

The BFDL section was added but certainly this could be changed if Easybuild is not maintained as a BFDL model. The other models are **Liberal Contribution** or **Meritocracy**. See https://opensource.guide/leadership-and-governance/

PR #591, opened 2020-02-04
urlchecker github action

shahzebsiddiqui:


shahzebsiddiqui:
@boegel this should work once approved and added to easybuilders project from marketplace https://github.com/marketplace/actions/urls-checker

ocaisa:
I tested this in https://github.com/easybuilders/easybuild/pull/597 and I didn't like it. It

ocaisa:
Sorry I didn't mean to do that!

ocaisa:
IMO this is way too slow, we need to be able to whitelist all of our repo links and only check external links. From the docs, it doesn't look like white-listing `https://github.com/easybuilders/easybuild-easyconfigs/*` is possible

ocaisa:
I tell a lie, it is possible...

ocaisa:
Note that this clones the repo and doesn't take a branch as an opton, so can only do something useful on a push to master.

Also, it can't deal with our templating (`%(version)s`) in the URLs, so there will always be manual work to keep a white-list up to date.

ocaisa:
FWIW Our templating is a real problem with this and the urlchecker crashes with some of our URLs:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/requests/models.py", line 379, in prepare_url
    scheme, auth, host, port, path, query, fragment = parse_url(url)
  File "/usr/local/lib/python3.8/site-packages/urllib3/util/url.py", line 392, in parse_url
    return six.raise_from(LocationParseError(source_url), None)
  File "<string>", line 3, in raise_from
urllib3.exceptions.LocationParseError: Failed to parse: http://%(namelower)s.googlecode.com/files
```

shahzebsiddiqui:
@ocaisa i have submitted a ticket to developer for feedback https://github.com/SuperKogito/URLs-checker/issues/7.

shahzebsiddiqui:
@ocaisa  even if this URL-checker may not be merged due to some of the issues. I think going through the report of all invalid URLs and fixing them first would be one major improvement. For end-user reading documentation and coming through a broken link can be problematic.

shahzebsiddiqui:
@boegel and @ocaisa the github action has changed its now https://github.com/urlstechie/urlchecker-action which is reimplementation of the one I sent you.

I think you can setup filter on what files to check or exclude. Anyhow I am happy to help resolve this old PR by refactoring this or we can close this PR.

PR #596, opened 2020-02-07
Add URL checker to CI

ocaisa:


boegel:
@ocaisa Why was this closed?

ocaisa:
I had added to the Black PR but it was too expensive and does not work on PRs, see https://github.com/easybuilders/easybuild/pull/591 which also covered the same ground

PR #597, opened 2020-02-07
Add GitHub Action tests

ocaisa:


.check_with_codespell.sh
```diff
@@ -0,0 +1 @@
+codespell --skip='.git,version-specific,scripts' --ignore-words-list=atleast,ninjs,simpy,proovread --quiet-level=2; retVal=$?; if [ $retVal -eq 1 ]; then echo 'Got one expected warning, so success!';  else $(exit $retVal); fi;
```
ocaisa: I  couldn't find a way to execute this line in a github action so I had to hide it in a script

.github/workflows/doc_build.yml
```diff
@@ -0,0 +1,35 @@
...
+      run: |
+        pip install flake8
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Test with codespell
+      run: |
+        pip install codespell
+        bash .check_with_codespell.sh
```
boegel: @ocaisa Would be better to move this script into `.github/workflows/`?

.github/workflows/doc_build.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Test with sphinx
+      run: |
+        pip install sphinx
+        READTHEDOCS=1 sphinx-build docs build
+    - name: Test installation
+      run: |
+        pip install .
```
boegel: add back comment `# test installation too (using options that don't require to have a modules tool installed)`

.github/workflows/doc_build.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Lint with flake8
+      run: |
+        pip install flake8
```
boegel: add missing comment that we have in `.travis.yml`?

.github/workflows/doc_build.yml
```diff
@@ -0,0 +1,35 @@
+name: Docs build
```
boegel: we check more than just docs, we also test the most recent `easybuild` installation end-to-end

boegel:
Thanks @ocaisa!

PR #600, opened 2020-02-18
add references to more information w.r.t. toolchains + move up link to common toolchains page on landing page

justbennet:
Per Slack discussion

What do you think of the idea of adding links to these two pages from

https://easybuild.readthedocs.io/en/latest/eb_list_toolchains.html
https://easybuild.readthedocs.io/en/latest/version-specific/toolchain_opts.html

along with some text that they are for more detailed information about toolchains and with a table of contents entry something like "More information about available toolchains"  to this page,

https://easybuild.readthedocs.io/en/latest/Common-toolchains.html

or in the Toolchains section of

https://easybuild.readthedocs.io/en/latest/Concepts_and_Terminology.html#toolchains

I was unable to find links to the two pages with more details via the table of contents, and I think that would help beginners who do not necessarily know the terms well to be able to find them from a section to which they are likely to browse via the ToC.


boegel:
Makes sense, thanks a lot @justbennet!

PR #602, opened 2020-02-23
fix marker for indicating what is installed in the output of -D

justbennet:
The current first example uses `[*]` in the `eb -Dr` output, whereas in the legend below the table, it says it will use [+] to indicate that the build is available, EasyBuild will skip building this module.  The other examples were updated to use the `[+]`.


boegel:
Nice catch @justbennet, thanks a lot for the fix!

PR #603, opened 2020-02-24
Update toolchain doc

justbennet:
I updated the example output in the Typical workflow example to use EB 4.1.1 and install WRF based on the `foss/2018b` toolchain.  References in the text have been changed to match the updated output.

Also changed some text in the table of toolchains from  docs/version-specific/eb_list_toolchains.txt to indicate there that the `dummy` toolchain is deprecated, and included text in the table note to refer the reader to the correct page for further information.

### DIFF
--- a/docs/Typical_workflow_example_with_WRF.rst
+++ b/docs/Typical_workflow_example_with_WRF.rst
@@ -26,30 +26,28 @@ Searching for build specification for a particular software package can be done 
 for example, to get a list of available easyconfig files for WRF::
 
   $ eb -S WRF
-  == temporary log file in case of crash /tmp/easybuild-MdAp7p/easybuild-zEBJMk.log
-  == Searching (case-insensitive) for 'WRF' in /home/example/.local/easybuild/software/EasyBuild/1.15.2/lib/python2.7/site-packages/easybuild_easyconfigs-1.15.2.0-py2.7.egg/easybuild/easyconfigs
-  CFGS1=/home/example/.local/easybuild/software/EasyBuild/1.15.2/lib/python2.7/site-packages/easybuild_easyconfigs-1.15.2.0-py2.7.egg/easybuild/easyconfigs/w/WRF
-   * $CFGS1/WRF-3.3.1-goalf-1.1.0-no-OFED-dmpar.eb
-   * $CFGS1/WRF-3.3.1-goolf-1.4.10-dmpar.eb
-   [...]
-   * $CFGS1/WRF-3.5-goolf-1.4.10-dmpar.eb
-   * $CFGS1/WRF-3.5-ictce-4.1.13-dmpar.eb
-   * $CFGS1/WRF-3.5-ictce-5.3.0-dmpar.eb
-   * $CFGS1/WRF-3.5.1-goolf-1.4.10-dmpar.eb
-   * $CFGS1/WRF-3.5.1-goolf-1.5.14-dmpar.eb
-   [...]
-  == temporary log file /tmp/easybuild-MdAp7p/easybuild-zEBJMk.log has been removed.
-  == temporary directory /tmp/easybuild-MdAp7p has been removed.
+  CFGS1=/home/bennet/.local/easybuild/software/EasyBuild/4.1.1/easybuild/easyconfigs
+   * $CFGS1/w/WPS/WPS-4.0.1_find-wrfdir.patch
+   * $CFGS1/w/WPS/WPS-4.0.2_find-wrfdir.patch
+   [ . . . ]
+   * $CFGS1/w/WRF/WRF-4.0.1-intel-2018b-dmpar.eb
+   * $CFGS1/w/WRF/WRF-4.0.2-foss-2018b-dmpar.eb
+   [ . . . ]
+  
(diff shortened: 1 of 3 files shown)


docs/version-specific/eb_list_toolchains.txt
```diff
@@ -14,7 +14,7 @@ cgmvapich2           Clang/GCC      MVAPICH2
 cgmvolf              Clang/GCC      MVAPICH2    OpenBLAS/LAPACK, ScaLAPACK(/BLACS), FFTW
 cgompi               Clang/GCC      OpenMPI
 cgoolf               Clang/GCC      OpenMPI     OpenBLAS/LAPACK, ScaLAPACK(/BLACS), FFTW
-dummy
+dummy                                           (Deprecated in favor of system toolchain)
```
boegel: @justbennet All files in `docs/version-specific` are re-generated using `docs/scripts/update-all-docs.sh`, so this will be overwritten again.

You should change the script instead to inject this?

docs/version-specific/eb_list_toolchains.txt
```diff
@@ -38,5 +38,6 @@ intel                icc, ifort     impi        imkl
 iomkl                icc, ifort     OpenMPI     imkl
 iqacml               icc, ifort     QLogicMPI   ACML, ScaLAPACK(/BLACS), FFTW
 ismkl                icc, ifort     MPICH2      imkl
+system                                          (See the note below)
```
boegel: What note below?

docs/Typical_workflow_example_with_WRF.rst
```diff
@@ -26,30 +26,28 @@ Searching for build specification for a particular software package can be done
...
-   * $CFGS1/WRF-3.5.1-goolf-1.5.14-dmpar.eb
-   [...]
-  == temporary log file /tmp/easybuild-MdAp7p/easybuild-zEBJMk.log has been removed.
-  == temporary directory /tmp/easybuild-MdAp7p has been removed.
-
-Various easyconfig files are found: for different versions of WRF (e.g., v3.5 and v3.5.1),
-for different (versions of) compiler toolchains (e.g., goolf v1.4.10, goolf v1.5.14, ictce), etc.
-
-For the remainder of this example, we will use the available ``WRF-3.5.1-goolf-1.4.10-dmpar.eb``
+  CFGS1=/home/bennet/.local/easybuild/software/EasyBuild/4.1.1/easybuild/easyconfigs
```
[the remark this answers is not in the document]
justbennet: Oh, right!  :man_facepalming: Yes, I'll go change my name to `example`.

docs/version-specific/eb_list_toolchains.txt
justbennet: This table comes from an included file, so the note is actually in the file that includes this, which is docs/eb_list_toolchains.rst.  The note is added in the change request to that.

docs/version-specific/eb_list_toolchains.txt
justbennet: OK.  I'll look at the script.  Sorry about that.

justbennet:
@boegel Sorry, my lack of Git/GitHub foo is showing.  I updated the branch in my fork that is the source for this PR, but the changes made there seem not to have propagated here.  I thought from a previous mistake that they would have done.

Also, I am not seeing a script that regenerates docs/version-specific/eb_list_toolchains.txt.  The `script docs/scripts/update-all-docs.sh` seems only to generate `.rst` files.

```
$ find docs/scripts -type f -exec grep -H toolchains.rst {} \;
docs/scripts/update-all-docs.sh:echo ".. _vsd_list_toolchains:" > $VERSION_SPECIFIC_DIR/toolchains.rst
docs/scripts/update-all-docs.sh:echo >> $VERSION_SPECIFIC_DIR/toolchains.rst
docs/scripts/update-all-docs.sh:echo "*(see also* \`\`eb --list-toolchains\`\` *)*" >> $VERSION_SPECIFIC_DIR/toolchains.rst
docs/scripts/update-all-docs.sh:echo >> $VERSION_SPECIFIC_DIR/toolchains.rst
docs/scripts/update-all-docs.sh:eb --list-toolchains --output-format rst >> $VERSION_SPECIFIC_DIR/toolchains.rst
```
The only references to `.txt` are
```
$ find docs/scripts -type f -exec grep -H '.txt' {} \;
docs/scripts/gen_easyblocks_docs.py:    write_file(so.options.out_file, txt)
docs/scripts/gen_easyblocks_docs.py:    print(txt)
```

Sorry to be stumble-footed about this.

It might be easier if I make the corrections, close this PR as unfulfilled, and open two new, separate PRs, one for the WRF changes and one for the toolchain list changes?

Sorry to put this out while you're on leave.

justbennet:
I am closing this PR.  The WRF examples have been resubmitted as PR #604  

I will return to the other after investigating/inquiring about the script generation of the pages involved.

boegel:
@justbennet No need to apologize at all, I really appreciate you contributing to the docs!

In fact, it seems like you've found a bug! `version-specific/eb_list_toolchains.txt` isn't getting updated at all, which is why you can't find a reference to it in the script. We should get that fixed, obviously...

I guess we should be using the output of `eb --list-toolchains --output-format=rst` for this?

Please open an issue for this to report it.

PR #604, opened 2020-02-24
Updating WRF examples with output from EB 4.1.1 and foss/2018b

justbennet:
This is separating out the changes for WRF example from other documentation
changes that were accidentally bundled in PR #603.  Repeating the summary
here.  This should supercede PR #603, which should be declined or closed.  If you would
like me to do so, I surely can, just let me know.

I updated the example output in the Typical workflow example to use EB
4.1.1 and install WRF based on the foss/2018b toolchain. References in
the text have been changed to match the updated output.

Sorry for the confusion and any extra work.

Addressed in this from #603, username changed to `example` everywhere.

### DIFF
--- a/docs/Typical_workflow_example_with_WRF.rst
+++ b/docs/Typical_workflow_example_with_WRF.rst
@@ -38,12 +38,12 @@ for example, to get a list of available easyconfig files for WRF::
 
 
 
-Various easyconfig files are found: for different versions of WRF (e.g., v3.8 and v4.1.3),
+Various easyconfig files are found: for different versions of WRF (e.g., v4.0.1 and v4.4.0.2),
 for different (versions of) compiler toolchains (e.g., foss 2018b, intel 2018b), etc.
 
 For the remainder of this example, we will use the available ``WRF-4.0.2-foss-2018b-dmpar.eb``
 easyconfig file to specify to EasyBuild to build and install
-WRF v4.0.2 using version 2018b of the ``foss`` toolchain.
+WRF v4.0.2 using version 2018b of the ``foss`` toolchain, which is one of the :ref:`common_toolchains`.
 
 See :ref:`toolchains_table` for a list of available toolchains. The ``foss`` toolchain
 stands for ``GCC, OpenMPI, OpenBLAS/LAPACK, ScaLAPACK, and FFTW``.  The ``foss`` toolchain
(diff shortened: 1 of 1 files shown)


docs/Typical_workflow_example_with_WRF.rst
```diff
@@ -26,30 +26,28 @@ Searching for build specification for a particular software package can be done
...
+   [ . . . ]
+   * $CFGS1/w/WRF/WRF-4.0.1-intel-2018b-dmpar.eb
+   * $CFGS1/w/WRF/WRF-4.0.2-foss-2018b-dmpar.eb
+   [ . . . ]
+  
+  Note: 16 matching archived easyconfig(s) found, use --consider-archived-easyconfigs to see them
+
+
+
+Various easyconfig files are found: for different versions of WRF (e.g., v3.8 and v4.1.3),
```
boegel: @justbennet I would mention versions here that are shown in the output above? So `v4.0.1` and `v4.0.2`?

docs/Typical_workflow_example_with_WRF.rst
```diff
@@ -26,30 +26,28 @@ Searching for build specification for a particular software package can be done
...
+
+For the remainder of this example, we will use the available ``WRF-4.0.2-foss-2018b-dmpar.eb``
 easyconfig file to specify to EasyBuild to build and install
-WRF v3.5.1 using version 1.4.10 of the ``goolf`` toolchain.
+WRF v4.0.2 using version 2018b of the ``foss`` toolchain.
 
-See :ref:`toolchains_table` for a list of available toolchains. The ``goolf`` toolchain
-stands for ``GCC, OpenMPI, OpenBLAS, ScaLAPACK and FFTW``.
+See :ref:`toolchains_table` for a list of available toolchains. The ``foss`` toolchain
+stands for ``GCC, OpenMPI, OpenBLAS/LAPACK, ScaLAPACK, and FFTW``.  The ``foss`` toolchain
```
boegel:
```
, and is one of the :ref:`common_toolchains`.
```

docs/Typical_workflow_example_with_WRF.rst
justbennet: I believe I made this change in the second commit of this PR.

docs/Typical_workflow_example_with_WRF.rst
justbennet: I believe I addressed this in the second commit of this PR.

PR #605, opened 2020-02-27
Update Lmod installation instructions

Flamefire:
Update versions and links used. Also link to Lmod docu and Github releases


ocaisa:
Apologies @Flamefire I missed this PR when mergeing #696, rather than ask you to update this again I created #780 (which incorporates this PR)

PR #615, opened 2020-04-14
release EasyBuild v4.2.0

migueldiascosta:


boegel:
Note:

* originally, the `easybuild-4.2.0` release was accidentally published to PyPI (on Tue April 15th 2020 at 2:28pm UTC) with an incorrect `setup.py` file, which specified `easybuild-easyblocks==4.1.1` and `easybuild-easdyconfigs==4.1.1` in `requires` and `install_requires`, see https://github.com/easybuilders/easybuild/blob/80c981c210b929271294f490d3ac4eded45e6a56/setup.py#L39 and https://github.com/easybuilders/easybuild/blob/80c981c210b929271294f490d3ac4eded45e6a56/setup.py#L82-L88

* hardcoding `easyblocks` and `easyconfigs` was done during the quick release of EasyBuild v4.1.2 (cfr. https://github.com/easybuilders/easybuild/pull/609/files#diff-2eeaed663bd0d25b7e608891384b7298);

* `setup.py` for EasyBuild v4.2.0 was fixed in 3a8c81b, shortly after `easybuild-4.2.0` was published on PyPI;

* a new release of the `easybuild` package on PyPI was published as [`easybuild-4.2.0.post0`](https://pypi.org/project/easybuild/4.2.0.post0) (since re-releasing under the exact same version is not allowed by PyPI, rightfully so), using the fixed `setup.py` from 7f8f08f;

* shortly after [`easybuild-4.2.0.post0`](https://pypi.org/project/easybuild/4.2.0.post0) was released to PyPI & tested, the faulty `easybuild-4.2.0` package was removed (on Tue April 15th 2020 at 3:50pm UTC), to avoid confusion & faulty installations of EasyBuild v4.2.0 (it was only available on PyPI for 1h22m);

* the bootstrap installation of EasyBuild was re-tested after `easybuild-4.2.0.post0` was published and after the `easybuild-4.2.0` package was removed, and found to be working as expected;

* this faulty release of `easybuild-4.2.0` could have only been picked up with *direct* installations from PyPI (using `pip install easybuild`), or when running the `bootstrap__eb.py` script when the faulty package was still available (between 2.28pm and 3.50pm (UTC) on Tue April 15th 2020)

PR #617, opened 2020-04-17
Add sections explaining exts_list options, and self.cfg in EasyBlocks

kelseymh:
Following the discussion in [EasyBlocks #3266](https://github.com/easybuilders/easybuild-framework/issues/3266), I've added a paragraph of text under "extensions parameters" explaining what each `exts_list` entry contains, and cross-referencing the "build parameters" section of the file.

I tried to use `rst2html` (provided with Python/3.6.6-foss-2018b) to check my work, but it apparently does not process `:ref:` directives (e.g., `rst2html Implementing-easyblocks.rst` generated a bunch of error messages).

Updated 25 Apr:  I am also adding a section to `Implementing-easyblocks.rst` as well, describing how to use the `self.cfg` dictionary, per #618 .


kelseymh:
This should **not** be merged as it is.  It appears that not all of the "build parameters" can be used directly in `exts_list` entries, or else I'm just not doing it properly.  Either way, it's going to need more than a bare sentence to explain how.

boegel:
@kelseymh The documentation under the `version-specific` directory is automatically regenerated with every EasyBuild release (using the `update-all-docs.sh` script in `docs/scripts`), so your additions will get trashed the way they're done currently...

Perhaps it's better to add a subsection in https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html instead?

For which parameters are you seeing problems?

There indeed are a couple of exceptions, one is `sources` (you need to use `source_tmpl` instead, which implies that you can only have a single source per extension currently), but that's one we actually want to fix so `sources` also works as expected for extensions...

kelseymh:
@boegel wrote:
> The documentation under the `version-specific` directory is automatically regenerated with every EasyBuild release (using the update-all-docs.sh script in docs/scripts), so your additions will get trashed the way they're done currently...

Oops!  Hence the directory name :-/  Okay, I can definitely back that out.

> Perhaps it's better to add a subsection in https://easybuild.readthedocs.io/en/latest/Writing_easyconfig_files.html instead?

That works for me.

> For which parameters are you seeing problems?
>
> There indeed are a couple of exceptions, one is `sources` (you need to use `source_tmpl` instead, which implies that you can only have a single source per extension currently), but that's one we actually want to fix so `sources` also works as expected for extensions...

I was having trouble getting the `filename`+`git_config` substructure to work properly, but that's probably due to the `sources`/`source_tmpl` mismatch you just mentioned.  Once I get this bit to work, then I'll have more confidence writing up guidance for others.

kelseymh:
Added '(WIP)' to title to reflect that the organization of my changes needs to be modified per @boegel .

docs/Implementing-easyblocks.rst
```diff
@@ -254,6 +254,37 @@ redefining it entirely, since we call out to the original ``configure_step`` met
...
+
+you could access these within an easyblock via
+
+.. code:: python
+
+  ec_name = self.cfg['name']
+  ec_ver = self.cfg['version']
+  ec_suffix = self.cfg['versionsuffix']
+
+For these particular "universal" arguments, this mapping is done for you in the ``EasyBlock`` base
```
ocaisa, line 280: I think what you mean here is that `self.name` is a valid class variable, right? As this stands it's a little confusing  because  if I have a custom variable `myvar`, I won't have `self.myvar` by default but I will always have self.cfg['myvar'] and I could then set that class variable in the constructor.

If that is what you mean, rather than use `ec_name` use `self.name` in the assignments to make this a little clearer.

kelseymh: It _is_ confusing, not just as I wrote it here, but to a new developer trying to write an EasyBlock for the first time :-).   I've tried expanding the wording in this new section, and included a one-line usage example.

docs/Writing_easyconfig_files.rst
```diff
@@ -506,12 +506,86 @@ which is equivalent to::
...
+
+Since EasyBuild v4.2.1, the ``git_config`` dictionary (see :ref:`downloading-from-a-git-repository`)
+may be included in the ``exts_list`` source definition.  For example, to download Git sources from a
+private repository and convert them to a tar-ball for installation:
+
+.. code:: python
+
+  exts_defaultclass = 'PythonPackage'
+  exts_list = [
+      ('pyCAP', 'master', {
```
ocaisa, line 575: It's a detail, but I'd personally prefer to see a different tag as the `version` here, the example is going to give a non-reproducible build. How about:
```
    ('pyCAP', '0.1', {
          'source_tmpl': '%(name)s-%(version)s.tar.gz',
          'git_config': {
              'url': 'ssh://nero.stanford.edu/data/git/Analysis',
              'repo_name': 'pyCAP',
              'tag': 'v%(version)s',
          }
      }),
```

docs/Implementing-easyblocks.rst
ocaisa: That part looks good to me now

docs/Writing_easyconfig_files.rst
kelseymh: Point taken.  I can give it as an example, but that particular (private!) package doesn't have tags (we're slowly trying to migrate away from using it at all).

kelseymh:
@ocaisa One question for you -- should these documentation updates be rolled out before the new feature in [Framework PR #3294](https://github.com/easybuilders/easybuild-framework/pull/3294)?

docs/Writing_easyconfig_files.rst
```diff
@@ -506,12 +506,86 @@ which is equivalent to::
...
+.. code:: python
+
+  exts_defaultclass = 'PythonPackage'
+  exts_list = [
+      ('pyCAP', '0.1', {
+          'source_tmpl': '%(name)s-%(version)s.tar.gz',
+          'git_config': {
+              'url': 'ssh://nero.stanford.edu/data/git/Analysis',
+              'repo_name': 'pyCAP',
+              'tag': 'master',
```
ocaisa, line 580: This one should change too, `v%(version)s`?

ocaisa:
@kelseymh I haven't been close enough attention but of course you're right, https://github.com/easybuilders/easybuild-framework/pull/3294 should come first.

kelseymh:
How do I deal with the "codespell" failure on the "Docs build and installation" test?  

https://github.com/easybuilders/easybuild/pull/617/checks?check_run_id=754862987#step:5:1

ocaisa:
Fix the spelling error at https://github.com/easybuilders/easybuild/pull/617/checks?check_run_id=754862987#step:5:17 and add `namd` to the ignore words list in `easybuild/.github/workflows/check_with_codespell.sh`

ocaisa:
@kelseymh argh sorry, Travis ci needs the ignore word as well

ocaisa:
https://github.com/easybuilders/easybuild/blob/develop/.travis.yml#L16

kelseymh:
@ocaisa @boegel This PR is still listing "one change requested", but I can't figure out what the change is.  Following the "See review" takes me to @ocaisa 's comment about using a tag, rather than a branch or SHA in my example, which I implemented.  I'm hoping this could get re-reviewed with the current version so it can roll out to 4.2.2 along with the new code it's documenting.

docs/Implementing-easyblocks.rst
```diff
@@ -254,6 +254,39 @@ redefining it entirely, since we call out to the original ``configure_step`` met
...
+  version = '2.5.3'
+  versionsuffix = '-Python-3.7.4'
+
+these are accessible within an easyblock via
+
+.. code:: python
+
+  longform = ''.join(self.cfg['name'],'/',self.cfg['version'],self.cfg['versionsuffix'])
+
+For these "universal" arguments, this mapping is done for you in the ``EasyBlock`` base class, so
```
ocaisa, line 278: `For these "universal" arguments, an additional mapping is done for you in the ``EasyBlock`` base class, so`

docs/Implementing-easyblocks.rst
```diff
@@ -254,6 +254,39 @@ redefining it entirely, since we call out to the original ``configure_step`` met
...
+they are also available as local Python variables (``self.name``, ``self.version``, etc.):
```
ocaisa, line 279:
```
they are also available as local Python variables (``self.name``, ``self.version``, etc.). Our previous example can therefore also be written:
.. code:: python
  longform_name = ''.join(self.name, '/', self.version, self.versionsuffix)
```

docs/Writing_easyconfig_files.rst
```diff
@@ -510,12 +510,87 @@ which is equivalent to::
...
+.. code:: python
+
+  exts_list = [
+      ('name', 'version', { 'option':'value', 'option':'value' })
+  ]
+
+The latter may contain essentially any of the full EasyConfig parameters, including ``buildopts``, ``installopts``, etc.
+Among those options, the following exceptions and special cases should be noted:
+
+* **nosource**: If set ``True``, no download will be done
```
ocaisa, line 539: How about
```
***``nosource``**
```
to make it totally clear that these are keys (same for the others below)

docs/Writing_easyconfig_files.rst
```diff
@@ -510,12 +510,87 @@ which is equivalent to::
...
+          'source_urls': ['https://pypi.python.org/packages/source/s/singledispatch/'],
+          'checksums': ['5b06af87df13818d14f08a028e42f566640aef80805c3b50c5056b086e3c2b9c'],
+      }),
+      (name, version, {
+          'source_urls': ['https://pypi.python.org/packages/source/n/numba/'],
+          'checksums': ['c62121b2d384d8b4d244ef26c1cf8bb5cb819278a80b893bf41918ad6d391258'],
+      }),
+  ]
+
+That third instance uses the ``name`` and ``version`` variables defined in the EasyConfigs file.  Since EasyBuild
```
ocaisa, line 569: `EasyConfigs` -> `easyconfig`, I think we use it that way pretty consistently

docs/Writing_easyconfig_files.rst
```diff
@@ -510,12 +510,87 @@ which is equivalent to::
...
+                  'url': 'ssh://nero.stanford.edu/data/git/Analysis',
+                  'repo_name': 'pyCAP',
+                  'tag': '%(version)s',
+              }
+          }
+      }),
+  ]
+
+Here, the template strings ``%(name)s`` and ``%(version)s`` will be substituted from the
+``exts_list`` entry elements ("pyCAP" and "0.1", respectively), not from the EasyConfigs values.
```
ocaisa, line 591: `EasyConfigs` -> `easyconfig`

docs/Writing_easyconfig_files.rst
kelseymh: I was following the same notation used in the bullet list under "Mandatory easyconfig parameters," and elsewhere (in fact, _every_ bullet list of parameters uses bold, but not text).  

If you think we should change them all, I can do that, but all of the parameter lists should be done consistently.  Please reply if you'll like me to change, otherwise I'm going to leave this alone.

docs/Writing_easyconfig_files.rst
kelseymh: I did a global search and replace.  I had used camel-case frequently; I've switched to lowercase.

docs/Writing_easyconfig_files.rst
ocaisa: You're right, consistency is king, this is fine as is

docs/Implementing-easyblocks.rst
```diff
@@ -254,6 +254,39 @@ redefining it entirely, since we call out to the original ``configure_step`` met
...
+
+  name = 'example'
+  version = '2.5.3'
+  versionsuffix = '-Python-3.7.4'
+
+these are accessible within an easyblock via
+
+.. code:: python
+
+  longform = ''.join(self.cfg['name'],'/',self.cfg['version'],self.cfg['versionsuffix'])
```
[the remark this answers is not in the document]
kelseymh: I think because I rewrote the text around it, this doesn't show up as "Outdated."

docs/Implementing-easyblocks.rst
```diff
@@ -254,6 +254,44 @@ redefining it entirely, since we call out to the original ``configure_step`` met
...
+
+You can use this notation successfully in your easyblock.  For all of the variables listed
+:ref:`vsd_avail_easyconfig_params`, the ``EasyBlock`` base class copies those dictionary entries into local Python
+variables:
+
+* ``self.name = self.cfg['name']``
+* ``self.version = self.cfg['version']``
+* ``self.versionsuffix = self.cfg['versionsuffix']``
```
boegel, line 283: @kelseymh This is not correct actually...

`self.name` and `self.version` are special cases, see https://github.com/easybuilders/easybuild-framework/blob/d66c0e518d36a614e79c75fed73ec60d94b77f32/easybuild/framework/easyconfig/easyconfig.py#L1084-L1096 .

There is no `self.versionsuffix`, and in general you need to use `self.cfg['name_of_parameter']`.

docs/Writing_easyconfig_files.rst
```diff
@@ -510,12 +510,87 @@ which is equivalent to::
...
 
+
 Using external modules as dependencies
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Since EasyBuild v2.1, specifying modules that are not provided via EasyBuild as dependencies is also supported.
 See :ref:`using_external_modules` for more information.
 
+
+Module Extensions
```
boegel, line 521: Just "Extensions" is better, since there's not really extensions to the module file, but extensions to the software provided by the module... :)

docs/Implementing-easyblocks.rst
kelseymh: I'll get rid of self.versionsuffix, but it seems like there are other cases than just `self.name` and `self.version`.  Near the code you cited, I see `@property def toolchain(self)` and `@property def all_dependencies(self)`.

I continue to learn about Python.  The `@property` keyword is what allows these things to be reference without a function call (i.e., `self.name` instead of `self.name()`.  I'll update this to list just the four `@property` cases, and make it clear those are special.

kelseymh: Updates are in PR #634

PR #629, opened 2020-05-20
add definitions for foss/2020a + intel/2020a common toolchains to docs

boegel:


migueldiascosta:
Going in, thanks @boegel!

PR #634, opened 2020-07-03
Clarify recent self.cfg and extensions documentation

kelseymh:
Explicitly list easyblock 'properties' (which look like data member names) which reference `self.cfg` dictionary entries, and how to use them.  Don't use 'module' qualifier for extensions.

See https://github.com/easybuilders/easybuild/pull/617/files#r449669160


boegel:
Thanks @kelseymh!

PR #652, opened 2020-10-23
Fix broken create index command

bedroge:
I tried to run the following example from this page, but it didn't work:

```
$ eb --create-index --index-max-age=0 $HOME/easybuild/cit-hpc-easybuild/easyconfigs
Usage: eb [options] easyconfig [...]

eb: error: Value '--index-max-age=0' starts with a '-'. Use '--create-index=--index-max-age=0' if the value is correct.
```

This does work:
```
$ eb --index-max-age=0 --create-index $HOME/easybuild/cit-hpc-easybuild/easyconfigs
```

And, according to `eb --help`, it should in principle even be `--create-index=$HOME/easybuild/cit-hpc-easybuild/easyconfigs` (in that case, more commands on this page should be fixed)?

### DIFF
--- a/docs/Easyconfigs_index.rst
+++ b/docs/Easyconfigs_index.rst
@@ -137,7 +137,7 @@ By default, EasyBuild will consider index files to remain valid for 1 week (7 * 
 
 To create an index that *always* remains valid (never expires), use zero (``0``) as value for ``--index-max-age``::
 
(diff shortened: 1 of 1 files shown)


boegel (approved):

boegel:
The `=` is always optional, space can also be used as a separator, but the order of arguments is indeed wrong here, thanks for the fix @bedroge!

PR #673, opened 2021-02-07
Prevent wrong dash appearing in doc

branfosj:
The previous becomes `–pr-commit-msg`. This makes it `--pr-commit-msg` and similarly for the other two.


boegel (approved):
lgtm

Thanks for hunting this down @branfosj !

boegel:
@branfosj Is it worth adding a check in the tests to avoid that we re-introduce issues like this?

PR #674, opened 2021-02-08
Adding conditional symlinking of lua posix lib in GH workflow script

justbennet:
The Ubuntu 20 that GitHub has started using has the `lua/5.2/posix.so` file (or link) in it already, so the script that creates it unconditionally for Ubuntu 18 fails because it exists.

Wrapped the `ln` command in an if so that it only tried to create the symlink if the target does not exist.

### DIFF
--- a/.github/workflows/scripts.yml
+++ b/.github/workflows/scripts.yml
@@ -21,7 +21,7 @@ jobs:
       run: |
         sudo apt-get install lua5.2 liblua5.2-dev lua-filesystem lua-posix tcl-dev
         # fix for lua-posix packaging issue, see https://bugs.launchpad.net/ubuntu/+source/lua-posix/+bug/1752082
-        # needed for 18.04 not for 20.04.  Conditional added to skip if posix.so exists
+        # needed for Ubuntu 18.04, but not for Ubuntu 20.04, so skipping symlinking if posix.so already exists
         if [ ! -e /usr/lib/x86_64-linux-gnu/lua/5.2/posix.so ] ; then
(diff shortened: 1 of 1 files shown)


Flamefire (approved):

.github/workflows/scripts.yml
```diff
@@ -21,7 +21,10 @@ jobs:
       run: |
         sudo apt-get install lua5.2 liblua5.2-dev lua-filesystem lua-posix tcl-dev
         # fix for lua-posix packaging issue, see https://bugs.launchpad.net/ubuntu/+source/lua-posix/+bug/1752082
-        sudo ln -s /usr/lib/x86_64-linux-gnu/lua/5.2/posix_c.so /usr/lib/x86_64-linux-gnu/lua/5.2/posix.so
+        # needed for 18.04 not for 20.04.  Conditional added to skip if posix.so exists
```
boegel, line 24: @justbennet Let's improve the comment a little bit:

```suggestion
        # needed for Ubuntu 18.04, but not for Ubuntu 20.04, so skipping symlinking if posix.so already exists

boegel (changes requested):

boegel:
We could use this same fix in the other repos too (`easybuild-framework`, `easybuild-easyblocks`, `easybuild-easyconfigs`), are you up for opening a PR for the same change there?

justbennet:
> We could use this same fix in the other repos too (`easybuild-framework`, `easybuild-easyblocks`, `easybuild-easyconfigs`), are you up for opening a PR for the same change there?

Oh, I could manage that, I think.  Won't even cost you a chit for a post-COVID beer.  ;-)

boegel (approved):
lgtm

boegel:
Thanks @justbennet!

PR #690, opened 2021-04-09
add EasyBuild v4.3.4 release notes

boegel:

PR #691, opened 2021-04-09
make sure that release notes are in place for current EasyBuild version

boegel:


boegel:
merged via #690, which shows that the test works as expected

PR #693, opened 2021-04-13
Change unified diff command

rpep:
Puzzled over this one for a while when following the current instructions (not being familiar with using diff and patch). 

If a patch requires removal of a file, removing the file + running 'diff -ru' is not sufficient - you need to also add the N flag to get it to treat the removed file as empty which makes it be removed when patch is called.


ocaisa (approved):
Thanks for the improvement

ocaisa:
@rpep Your PR targest `main` rather than `develop`, can you please switch the target branch (`main` is protected so I can't merge this)

rpep:
Apologies, done!

ocaisa (approved):

PR #714, opened 2021-06-10
Add documentation and example for write-module-hook

Flamefire:
Also fix some typos and minor issues

### DIFF
--- a/docs/Hooks.rst
+++ b/docs/Hooks.rst
@@ -59,12 +59,14 @@ Available hooks
 Currently (since EasyBuild v3.7.0), three types of hooks are supported:
 
 * ``start_hook`` and ``end_hook``, which are triggered *once* before starting software installations,
-  and *once* right after completing all installations, respectfully;
-* ``parse_hook``, which is triggered when an easyconfig file is being parsed;
+  and *once* right after completing all installations, respectively
+* ``parse_hook``, which is triggered when an easyconfig file is being parsed
+* ``module_write_hook``, which is triggered right before a module file is written.
+  This includes the temporary module files used for extensions, sanity checks and the devel module!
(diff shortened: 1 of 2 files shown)


boegel (approved):

docs/Hooks.rst
```diff
@@ -119,6 +123,12 @@ Do take into account the following:
 * for ``parse_hook``, one argument is provided: the ``EasyConfig`` instance
   that corresponds to the easyconfig file being parsed (usually referred to as ``ec``)
 
+* for ``module_write_hook``, 3 arguments are provided:
+   * the ``EasyBlock`` instance used to perform the installation (usually referred to as ``self``)
+   * the filepath of the module that will be written
+   * the module text as a string
+  The return value of this hook (if any) will then be appended to the contents of the module file.
```
Flamefire, line 130: @boegel You changed this sentence and introduced a mistake: The contents will be **written** to the module file. NOT appended, just as I wrote initially. The whole return value will be used and written directly, so this is misleading

Or what did you mean by "appended" here?

boegel: Oh, and I even checked the code before changing this... Will fix.

Flamefire: For rerference see https://github.com/easybuilders/easybuild-framework/blob/96977045496d8a5106a54727f6522c92863008e8/easybuild/framework/easyblock.py#L3168-L3170 and  https://github.com/easybuilders/easybuild-framework/blob/96977045496d8a5106a54727f6522c92863008e8/easybuild/framework/easyblock.py#L3179

boegel: Fixed in #715, thanks for double checking and catching this!

PR #715, opened 2021-06-14
correct statement on return value of module write hook

boegel:
fix incorrect statement I introduced in #714 🙊

### DIFF
--- a/docs/Hooks.rst
+++ b/docs/Hooks.rst
@@ -127,7 +127,7 @@ Do take into account the following:
    * the ``EasyBlock`` instance used to perform the installation (usually referred to as ``self``)
    * the filepath of the module that will be written
    * the module text as a string
-  The return value of this hook (if any) will then be appended to the contents of the module file.
+  The return value of this hook (if any) is used as contents for the module file that will be written.
 
 * for the step hooks, one argument is provided:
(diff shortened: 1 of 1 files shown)


akesandgren (approved):
LGTM

docs/Hooks.rst
```diff
@@ -127,7 +127,7 @@ Do take into account the following:
    * the ``EasyBlock`` instance used to perform the installation (usually referred to as ``self``)
    * the filepath of the module that will be written
    * the module text as a string
-  The return value of this hook (if any) will then be appended to the contents of the module file.
+  The return value of this hook (if any) is used as contents for the module file that will be written.
```
Flamefire, line 130: TBH: I find "as contents for the file that will be written" a bit to complicated compared to "will be written to the file"
Maybe we could still clarify this a bit, especially with respect to "no return value/no hook":

```suggestion
  The return value of this hook, when set, will replace the original text that is then written to the module file.
```

akesandgren:
Going in, thanks @boegel!

docs/Hooks.rst
boegel: That's better indeed, opening another PR :)

boegel: see #716

Issue #713, opened 2021-06-10
bootstrap_eb.py problem


Lueders-Martin:
Dear All,

I have some issues trying to install EasyBuild on my Debian 11 system:
I created a virtual environment with virtualenv 20.4.7 using
```
virtualenv -p python3 eb-generic
```
and in the created directory run:
```
. bin/activate
wget https://raw.githubusercontent.com/easybuilders/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py

python bootstrap_eb.py .
```
which crashes with:
```
[[INFO]] EasyBuild bootstrap script (version 20210106.01, MD5: 2e9a1c86bea34d12b2fab3a8e79932fd)
[[INFO]] Found Python 3.9.2 (default, Feb 28 2021, 17:03:44) ; [GCC 10.2.1 20210110]

[[INFO]] Installation prefix /home/lueders/eb-generic
[[INFO]] Found module command '/usr/share/lmod/lmod/libexec/lmod' via $LMOD_CMD (Lmod), so using it.
[[INFO]] Suitable setuptools installation already found, skipping stage 0...


[[INFO]] +++ STAGE 1: installing EasyBuild in temporary dir with easy_install...

[[INFO]] installing EasyBuild with 'easy_install --quiet --upgrade --prefix=/tmp/tmp0xbvou13/eb_stage1 easybuild>=4.0'

[[INFO]] Note: a 'SyntaxError' may be reported for the easybuild/tools/py2vs3/py2.py module.
You can safely ignore this message, it will not affect the functionality of the EasyBuild installation.

[[ERROR]] Running 'easy_install --quiet --upgrade --prefix=/tmp/tmp0xbvou13/eb_stage1 easybuild>=4.0' failed: module 'setuptools.command.easy_install' has no attribute 'main'
Traceback (most recent call last):
  File "/home/lueders/eb-generic/bootstrap_eb.py", line 380, in run_easy_install
    easy_install.main(args)
AttributeError: module 'setuptools.command.easy_install' has no attribute 'main'
```
Any idea what I am missing?

boegel:
@Lueders-Martin The bootstrap script hasn't received a lot of love recently, and it may not be compatible with recent `setuptools` versions anymore.

I strongly recommend to use `pip install easybuild`, especially in a virtualenv.

[...]

PR #718, opened 2021-06-16
revamp EasyBuild installation docs: recommend pip or (manual) 3-stage bootstrap with pip + EasyBuild

branfosj:
Change the install instructions to avoid the bootstrap method. See #713 for an example of why.

* pip install recommended
* also eb as a module install


boegel (changes requested):
Some tweaks worth including in https://github.com/branfosj/easybuild/pull/2

boegel (approved):

PR #780, opened 2022-02-04
Update installation instructions for Lmod

ocaisa:
Also incorporates PR #605

### DIFF
--- a/docs/Installing-Lmod-without-root-permissions.rst
+++ b/docs/Installing-Lmod-without-root-permissions.rst
@@ -7,18 +7,25 @@ Installing Lmod without root permissions
 This short guide will show how to install Lmod (and Lua, on which it
 depends) on Linux, without requiring root permissions.
 
+For full instructions on installing Lmod and its prerequisites check
+the `Lmod documentation<https://lmod.readthedocs.io/en/latest/030_installing.html>`_.
+
 Lua
 ~~~
 
-Dependencies: Installing Lua using the steps below requires rsync, make and gcc
+Dependencies: Installing Lua using the steps below requires bzip2, rsync, make and gcc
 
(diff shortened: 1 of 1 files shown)


docs/Installing-Lmod-without-root-permissions.rst
```diff
@@ -55,7 +62,8 @@ Lmod
 Dependencies: building Lmod using the steps below requires tcl, tcl-dev(el), make and bzip2
 
 **Step 1**: Download and unpack the latest available Lmod version,
-`Lmod-8.4.tar.bz2`_ at the time of writing.
+`Lmod-8.4.tar.bz2`_ at the time of writing, from the `Lmod GitHub releases page
```
boegel, line 65: @ocaisa This should be updated too, Lmod 8.4 is not latest at time of writing anymore ;)

boegel (changes requested):

docs/Installing-Lmod-without-root-permissions.rst
```diff
@@ -67,18 +75,25 @@ Dependencies: building Lmod using the steps below requires tcl, tcl-dev(el), mak
 
     ./configure --prefix=$HOME && make install
 
-**Step 3**: Update ``$PATH`` so ``lmod`` is available (put this in your
-``.bashrc``):
+**Step 3**: Update the ``$LMOD_CMD`` to point to the Lmod executable
+(put this in your ``.bashrc``):
+
+.. code:: bash
+    export LMOD_CMD=$HOME/lmod/8.4/libexec/lmod
```
boegel, line 82: This is set by the Lmod init script, no need to do this manually at all (and if `$LMOD_CMD` is set, no need to fiddle with `$PATH`, indeed)

boegel (changes requested):

PR #796, opened 2022-04-08
git_config tag option is for getting a tag

branfosj:
In https://github.com/easybuilders/easybuild-framework/pull/3795 we changed to not allowing the git download method to get a branch, as this causes issues when we there are branchs and tags with the same name.


boegel (approved):
lgtm

PR #825, opened 2022-09-27
Improve the description of using a tuple in `patches`

Flamefire:
Clarify that the level must be an integer and describe the alternative to use a string to specify the starting directory. Also some minor rewording.


ocaisa (approved):
LGTM

PR #828, opened 2022-10-17
Document GITHUB_RELEASE & GITHUB_LOWER_RELEASE

Flamefire:


boegel (approved):

boegel:
All docs in the `version-specific` subdirectory are automatically updated for every EasyBuild release, but no harm in adding these already.

PR #855, opened 2023-01-20
Fix indentation in Easystack examples

bedroge:
The examples with the additional `options` don't work, e.g. the first one from https://docs.easybuild.io/en/latest/Easystack-files.html#combining-command-line-options-with-options-in-an-easystack-file results in:
```
ERROR: Failed to parse easystack file: expected a dictionary with one key (the EasyConfig name), instead found keys: PyTorch-1.12.0-foss-2022a-CUDA-11.7.0.eb, options, see https://docs.easybuild.io/en/latest/Easystack-files.html for documentation.
```

Looking at an example from the tests at https://github.com/easybuilders/easybuild-framework/blob/develop/test/framework/easystacks/test_easystack_easyconfigs_opts.yaml, the additional options need more indentation, and the following does indeed work:
```
easyconfigs:
  - PyTorch-1.12.0-foss-2022a-CUDA-11.7.0.eb:
      options:
        from-pr: 15924
        debug: True
  - OpenFOAM-v2206-foss-2022a.eb:

```

### DIFF
--- a/docs/Easystack-files.rst
+++ b/docs/Easystack-files.rst
@@ -70,13 +70,13 @@ For example:
 
   easyconfigs:
     - PyTorch-1.12.0-foss-2022a-CUDA-11.7.0.eb:
-      options:
-        from-pr: 15924
-        debug: True
+        options:
+          from-pr: 15924
+          debug: True
     - Hypre-2.25.0-foss-2022a.eb:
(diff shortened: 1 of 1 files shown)


boegel (approved):
lgtm

PR #861, opened 2023-02-16
Add EUM23 to easybuild.io

branfosj:


branfosj:
Done via cabd60a and b38bb6c

PR #894, opened 2024-06-12
fix `create_source_tarball.sh` script to be compatible with recent setuptools versions

boegel:
Recent versions of setuptools (>= 69.0.3, not sure?) 'normalize' package names by replacing dashes with underscores, so we source tarballs like `easybuild_framework-4.9.2.tar.gz` instead of `easybuild-framework-4.9.2.tar.gz`.

The changes being made here accept both, just to stay compatible with older setuptools versions (or in case they change their mind again...)

edit: see also https://github.com/easybuilders/easybuild-easyblocks/pull/3358

### DIFF
--- a/scripts/create_source_tarball.sh
+++ b/scripts/create_source_tarball.sh
@@ -22,7 +22,7 @@ function success() {
 }
 
 function warning() {
-    echo -e "\033[31mWARNING: $1\033[0m" >&2
+    echo -e "\033[33mWARNING: $1\033[0m" >&2
     warnings="$warnings$1 "
 }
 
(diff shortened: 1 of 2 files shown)


scripts/create_source_tarball.sh
```diff
@@ -22,7 +22,7 @@ function success() {
 }
 
 function warning() {
-    echo -e "\033[31mWARNING: $1\033[0m" >&2
+    echo -e "\033[33mWARNING: $1\033[0m" >&2
```
boegel, line 25: This just makes warnings yellow rather than red, so they stand out from errors a bit better

Micket (approved):
lgtm

PR #922, opened 2025-08-05
Remove deprecated license classifier

APN-Pucky:
Gentoo QA complains:

```
 * Messages for package sys-cluster/easybuild-5.1.1:

 * QA Notice: setuptools warnings detected:
 * 
 * 	License classifiers are deprecated.
```

Since classifiers for licenses have been deprecated and the license is already set directly I think it can be removed from the classifier list.


boegel (approved):
lgtm

We should make the same change in `setup.py` for framework/easyblocks/easyconfigs.

I'll look into that since we're really close to the EasyBuild v5.1.2 release, and this is a trivial change...

